PR #6, opened 2018-02-25
Remove voice functionality

ianagbip1oti:


ianagbip1oti:
Preferred the more conservative approach of #7.

PR #149, opened 2018-05-21
Added a guild cache into the repository.

CoreyShupe:
Small cleanup (very small) of repository, and added a guild cache. Nothing too huge.


src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
...
+import io.reactivex.Maybe;
+import io.reactivex.Observable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Stack;
+
+public class GuildsFromEvents implements GuildRepository {
+
+  private final Map<Snowflake, Guild> guildCache;
+  private final Map<Snowflake, Channel> channelCache;
```
ianagbip1oti: Should we cache Channels here, or should there be a `ChannelRepository`?

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
...
+import com.github.princesslana.eriscasper.repository.GuildRepository;
+import io.reactivex.Maybe;
+import io.reactivex.Observable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Stack;
+
+public class GuildsFromEvents implements GuildRepository {
+
+  private final Map<Snowflake, Guild> guildCache;
```
ianagbip1oti: How does this work when there are multiple threads?

My preferred option is to do as `UsersFromEvents`, we maintain the state using the `scan` methods in Observable, just because it fits with the function/reactive/immutable style.

Alternatively, this should at least be changed to a `ConcurrentMap`.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
...
+
+  public GuildsFromEvents(Observable<Event> eventObservable) {
+    guildCache = new HashMap<>();
+    channelCache = new HashMap<>();
+    eventObservable
+        .ofType(GuildCreateEvent.class)
+        .map(GuildCreateEvent::unwrap)
+        .subscribe(
+            guild -> {
+              guildCache.putIfAbsent(guild.getId(), guild);
```
ianagbip1oti: Why are we using `putIfAbsent`? If we've received another `GuildCreateEvent` with the same id, wouldn't the data from the new event be more up to date?

src/test/java/com/github/princesslana/eriscasper/faker/DataFaker.java
```diff
@@ -47,6 +45,38 @@ public static User user() {
         .build();
   }
 
+  public static Guild guild() {
+    Snowflake id = DiscordFaker.snowflake();
+    return ImmutableGuild.builder()
+        .id(id)
+        .name("FakeGuild")
```
ianagbip1oti: Fields such of these should use fake rather than constant data as well. Check out what's available in http://dius.github.io/java-faker/apidocs/index.html

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
...
+        .subscribe(
+            guild -> {
+              Snowflake id = guild.getId();
+              guildCache.remove(id);
+              Stack<Snowflake> removingChannelStack = new Stack<>();
+              channelCache
+                  .entrySet()
+                  .stream()
+                  .filter(entry -> id.equals(entry.getValue().getGuildId().orElse(null)))
+                  .forEach(entry -> removingChannelStack.push(entry.getKey()));
```
ianagbip1oti: Why do we push onto the stack here, rather than just deleting here?

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
...
+        .ofType(ChannelDeleteEvent.class)
+        .map(ChannelDeleteEvent::unwrap)
+        .subscribe(channel -> channelCache.remove(channel.getId()));
+  }
+
+  @Override
+  public Maybe<Guild> getGuild(Snowflake id) {
+    if (id == null) return Maybe.empty();
+    Guild got = guildCache.get(id);
+    return got == null ? Maybe.empty() : Maybe.just(got);
```
ianagbip1oti: `Maybes.fromNullable`
https://github.com/princesslana/ErisCasper.java/blob/master/src/main/java/com/github/princesslana/eriscasper/rx/Maybes.java#L9

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
...
+  public Maybe<Channel> getChannel(Snowflake id) {
+    if (id == null) return Maybe.empty();
+    Channel got = channelCache.get(id);
+    return got == null ? Maybe.empty() : Maybe.just(got);
+  }
+
+  @Override
+  public Maybe<Guild> getGuildFromChannel(Snowflake channelId) {
+    if (channelId == null) return Maybe.empty();
+    Channel channel = getChannel(channelId).blockingGet();
```
ianagbip1oti: We want to avoid `blockingXXX` methods. Should be possible use `map` here instead.

src/main/java/com/github/princesslana/eriscasper/repository/GuildRepository.java
```diff
@@ -0,0 +1,16 @@
...
+import io.reactivex.Maybe;
+
+public interface GuildRepository {
+  Maybe<Guild> getGuild(Snowflake id);
+
+  Maybe<Channel> getChannel(Snowflake id);
+
+  Maybe<Guild> getGuildFromChannel(Snowflake channelId);
+
+  Maybe<Snowflake> getGuildIdFromChannelId(Snowflake channelId);
```
ianagbip1oti: What advantage does having this method provide over `getGuildFromChannel`?

src/main/java/com/github/princesslana/eriscasper/repository/GuildRepository.java
```diff
@@ -0,0 +1,16 @@
...
+import com.github.princesslana.eriscasper.data.resource.Channel;
+import com.github.princesslana.eriscasper.data.resource.Guild;
+import io.reactivex.Maybe;
+
+public interface GuildRepository {
+  Maybe<Guild> getGuild(Snowflake id);
+
+  Maybe<Channel> getChannel(Snowflake id);
+
+  Maybe<Guild> getGuildFromChannel(Snowflake channelId);
```
ianagbip1oti: Is the method better at this level, or would be it better on a 'Channels' or 'Guilds' class? (like https://github.com/princesslana/ErisCasper.java/blob/master/src/main/java/com/github/princesslana/eriscasper/data/Users.java)

Should expect the user, if they have a channel id, to fetch the `Channel`, then get the guild via `Channel#getGuildId`?

ianagbip1oti (changes requested):
~~Need to get spotless working for you :)~~

~~It uses google-java-format, for which there is an intellij plugin also: https://github.com/google/google-java-format#intellij~~

Never mind, it's passing CI with spotless.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
+package com.github.princesslana.eriscasper.repository.event;
+
+import com.github.princesslana.eriscasper.data.Snowflake;
+import com.github.princesslana.eriscasper.data.event.*;
```
ianagbip1oti: [style nit] Avoid wildcard imports

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,93 @@
...
+  @Override
+  public Maybe<Guild> getGuild(Snowflake id) {
+    if (id == null) return Maybe.empty();
+    Guild got = guildCache.get(id);
+    return got == null ? Maybe.empty() : Maybe.just(got);
+  }
+
+  @Override
+  public Maybe<Channel> getChannel(Snowflake id) {
+    if (id == null) return Maybe.empty();
```
ianagbip1oti: I'm happy to assume parameters are not null. If the parameter was optional, we'd make it `Optional`.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
CoreyShupe: ConcurrentModificationException, I suppose switching to concurrent hashmaps should fix that so I'll remove the stack and see if tests still pass.

src/main/java/com/github/princesslana/eriscasper/repository/GuildRepository.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.github.princesslana.eriscasper.data.Snowflake;
+import com.github.princesslana.eriscasper.data.resource.Channel;
+import com.github.princesslana.eriscasper.data.resource.Guild;
+import io.reactivex.Maybe;
+import io.reactivex.annotations.NonNull;
+import java.util.Optional;
+
+public interface GuildRepository {
+  Maybe<Guild> getGuild(Optional<Snowflake> id);
```
ianagbip1oti: In what cases do we expect a client to call this when they don't have an id?

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.stream.Collectors;
+
+public class GuildsFromEvents implements GuildRepository {
+
+  private final ConnectableObservable<ImmutableMap<Snowflake, Guild>> guildCache;
+  private final ConnectableObservable<ImmutableMap<Snowflake, Channel>> channelCache;
+
+  public GuildsFromEvents(Observable<Event> eventObservable) {
+    ImmutableMap<Snowflake, Channel> emptyChannelMap =
+        ImmutableMap.<Snowflake, Channel>builder().build();
```
ianagbip1oti: `ImmutableMap.of()`

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class GuildsFromEvents implements GuildRepository {
+
+  private final ConnectableObservable<ImmutableMap<Snowflake, Guild>> guildCache;
+  private final ConnectableObservable<ImmutableMap<Snowflake, Channel>> channelCache;
+
+  public GuildsFromEvents(Observable<Event> eventObservable) {
```
ianagbip1oti: This constructor feels large and complex. What could we do to clarify what it's doing?

src/main/java/com/github/princesslana/eriscasper/rx/Observables.java
```diff
@@ -0,0 +1,15 @@
+package com.github.princesslana.eriscasper.rx;
+
+import io.reactivex.Observable;
+import java.util.Optional;
+
+public class Observables {
```
ianagbip1oti: Are these used now?

ianagbip1oti (changes requested):

src/main/java/com/github/princesslana/eriscasper/rx/Observables.java
CoreyShupe: No, I stopped using them, they'll be removed in my next commit.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
CoreyShupe: changed naming to `events` same as the other repository

src/main/java/com/github/princesslana/eriscasper/repository/GuildRepository.java
CoreyShupe: the Channel object contains an `Optional<Snowflake>` for the guildId

ianagbip1oti: Sure, but does it make sense to be handling the empty case here, or should the caller do it?

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
ianagbip1oti: I mean the whole constructor, not the parameter name.

e.g., Is there stuff we can break out into methods? Is there some comment block needed to explain what the method is doing as a whole?

CoreyShupe: I'll break it up, I can move it into descriptive methods, you're right.

src/main/java/com/github/princesslana/eriscasper/repository/GuildRepository.java
CoreyShupe: You're right, I'll change it to a nonnull param

src/main/java/com/github/princesslana/eriscasper/action/Action.java
```diff
@@ -22,11 +22,11 @@
    */
   Optional<I> getData();
 
-  public static <O> Action<Void, O> of(Route<Void, O> route) {
+  static <O> Action<Void, O> of(Route<Void, O> route) {
```
ianagbip1oti, line 25: why is the scope changing here?

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Override
+  public Maybe<Guild> getGuild(@NonNull Snowflake id) {
+    return guildWatcher.cache().firstElement().flatMap(map -> Maybes.fromNullable(map.get(id)));
+  }
+
+  @Override
+  public Maybe<Channel> getChannel(@NonNull Snowflake id) {
+    return channelWatcher
+        .cache()
```
ianagbip1oti: Is cache needed here? The replay should take care of this

src/test/java/com/github/princesslana/eriscasper/faker/DataFaker.java
```diff
@@ -47,6 +54,38 @@ public static User user() {
...
+        .afkTimeout(0L)
+        .verificationLevel(0L)
+        .defaultMessageNotifications(0L)
+        .explicitContentFilter(0L)
+        .mfaLevel(0L)
+        .addChannels(guildChannel(id))
+        .build();
+  }
+
+  public static Channel channel() {
```
ianagbip1oti: I'd prefer to have one faker per resource, with individual tests changing the values if needed. So, either `channel` or `guildChannel` (but renamed to `channel`), but not both.

src/test/java/com/github/princesslana/eriscasper/repository/event/TestGuildsFromEvents.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Test
+  public void getGuild_whenAfterGuildEvent_shouldCacheGuild() {
+    Guild guild = DataFaker.guild();
+    Channel channel = guild.getChannels().get(0);
+    events.onNext(GuildCreateEvent.of(guild));
+    Guild n = DataFaker.guild();
+    events.onNext(GuildCreateEvent.of(n));
+    subject.getGuild(n.getId()).test().assertValue(n);
+    subject.getGuild(guild.getId()).test().assertValue(guild);
```
ianagbip1oti: Can we add a comment here as to why we're calling this many times?

Or, break it out into a second test `getGuild_whenGetGuildCalledTwice_shouldReturnSameGuild` ?

ianagbip1oti (approved):

Issue #155, opened 2018-05-24
GuildDeleteEvent clears GuildCache


ianagbip1oti:
A `GuildDeleteEvent` results in all guilds being deleted.

```java
  @Test
  public void getGuild_whenDeleted_shouldMaintainNonDeleted() {
    Guild guild1 = DataFaker.guild();
    Guild guild2 = DataFaker.guild();

    // Create both guilds
    events.onNext(GuildCreateEvent.of(guild1));
    events.onNext(GuildCreateEvent.of(guild2));
    subject.getGuild(guild1.getId()).test().assertValue(guild1);
    subject.getGuild(guild2.getId()).test().assertValue(guild2);

    // Delete guild1 only
    events.onNext(
        GuildDeleteEvent.of(ImmutableUnavailableGuild.builder().id(guild1.getId()).build()));
    subject.getGuild(guild1.getId()).test().assertComplete().assertNoValues();
    subject.getGuild(guild2.getId()).test().assertValue(guild2);
  }
```

@CoreyShupe

CoreyShupe:
weird, I'll look into why that's happening, it should maintain itself

CoreyShupe:
This can be closed ~

ianagbip1oti:
Which PR was it fixed in?

CoreyShupe:
https://github.com/princesslana/ErisCasper.java/pull/157

PR #157, opened 2018-05-25
Fixed issue regarding caching, then deleting not properly deleting. (cleanup)

CoreyShupe:
Large cleanup along with redundancy cleanup as well as a fix for a caching issue deleting. Addresses bug: https://github.com/princesslana/ErisCasper.java/issues/155

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
+++ b/src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
@@ -9,17 +9,77 @@ import com.github.princesslana.eriscasper.data.event.GuildDeleteEvent;
 import com.github.princesslana.eriscasper.data.resource.Channel;
 import com.github.princesslana.eriscasper.data.resource.Guild;
 import com.github.princesslana.eriscasper.data.resource.ImmutableChannel;
+import com.github.princesslana.eriscasper.data.resource.UnavailableGuild;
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -36,106 +96,43 @@ public GuildsFromEvents(Observable<Event> eventObservable) {
...
+            eventObservable
+                .ofType(ChannelDeleteEvent.class)
+                .map(ChannelDeleteEvent::unwrap)
+                .map(REMOVE_CHANNEL_FUNCTION))
+        .scan(ImmutableMap.<Snowflake, Channel>of(), (map, function) -> function.apply(map))
+        .doOnError(Throwable::printStackTrace)
         .replay(1);
   }
 
+  // Used due to avoiding redundant code
```
ianagbip1oti, line 126: redundant comment :)

ianagbip1oti (changes requested):
Should the test from the bug report be added to the tests?

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
CoreyShupe: I added in code to the guild and channel test to test what the bug report pointed out. If you'd like I can add it in, also I'll remove the second comment

ianagbip1oti (approved):

Issue #130, opened 2018-03-29
Support Webhook Resource


ianagbip1oti:
https://discordapp.com/developers/docs/resources/webhook#webhook-resource

- [x] Create Webhook
- [x] Get Channel Webhooks
- [x] Get Guild Webhooks
- [x] Get Webhook
- [x] Get Webhook with Token
- [x] Modify Webhook
- [x] Modify Webhook with Token
- [x] Delete Webhook
- [x] Delete Webhook with Token
- [ ] Execute Webhook
- [ ] Execute Slack-Compatible Webhook
- [ ] Execute GitHub-Compatible Webhook

Issue #129, opened 2018-03-29
Support Voice Resource


ianagbip1oti:
https://discordapp.com/developers/docs/resources/voice#voice-resource

- [x] List Voice Regions

PR #158, opened 2018-06-03
Route changes / Webhook route addition.

CoreyShupe:
Added webhook resource along with changing some of the route options to fit other needs.
Addresses https://github.com/princesslana/ErisCasper.java/issues/130
Also addresses (https://github.com/princesslana/ErisCasper.java/issues/129) due to branching issues (whoops)
Adds new utilities for certain things
Updates ErisCasper to 0.5.0


CoreyShupe:
Yeh looks like we need to release another vers of Data before PRing this...

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
```diff
@@ -31,7 +31,9 @@
     GET("GET", Content.QUERY_STRING),
     PATCH("PATCH", Content.BODY),
     POST("POST", Content.BODY),
-    PUT("PUT", Content.BODY);
+    QUERY_POST("POST", Content.QUERY_STRING),
+    PUT("PUT", Content.BODY),
+    PUT_QUERY("PUT", Content.QUERY_STRING);
```
ianagbip1oti: With QUERY_POST or PUT_QUERY (these names are inconsistent ;)), how do I also send the body with that request?

src/main/java/com/github/princesslana/eriscasper/rest/webhook/ExecuteGithubWebhookRequest.java
```diff
@@ -0,0 +1,19 @@
...
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/webhook#execute-githubcompatible-webhook-querystring-params">
+ *     https://discordapp.com/developers/docs/resources/webhook#execute-githubcompatible-webhook-querystring-params</a>
+ */
+@Value.Immutable
+public interface ExecuteGithubWebhookRequest {
+  Optional<Boolean> isWait();
+
+  default String toQueryString() {
+    return new QueryStringBuilder().addBool("wait", isWait()).build();
```
ianagbip1oti: I don't see a `addBool` method added to query string (you may have missed including that change here), but let's not abbreviate inconsistently. I'd prefer `addBoolean`

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
```diff
@@ -157,13 +159,36 @@ public int hashCode() {
     return get(path, noContent(), jsonResponse(rsClass));
   }
 
+  public static <Rq, Rs> Route<Rq, Rs> get(
+      String path, Class<Rq> rqClass, Function<Response, Rs> rsFunction) {
```
ianagbip1oti: I'm not excited at all in introducing every possible combo of Function/Class parameters that can be here. I'd rather have two: Class/Class, Function/Function. If the definition has to use a Function for one, having to call jsonRequestBody() there I think is ok.

ianagbip1oti (changes requested):

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
CoreyShupe: We should make those kinds of methods public then, because they're very useful.

src/main/java/com/github/princesslana/eriscasper/rest/webhook/ExecuteGithubWebhookRequest.java
CoreyShupe: I need to add that into the branch, forgot to add the utility upgrades to this branch.

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
CoreyShupe: I need to rename them yes. Also, it's a different form of request, some of the API stuff need query strings in those places rather than body statements so I just added the ability to use that type of formatting. Not sure if there's a better way to do it.

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
ianagbip1oti: Yeah, I'm fine with that. I had been making them public as they were needed.

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
ianagbip1oti: But they don't need just a query string right? Doesn't something like execute slack webhook also need a request body?

CoreyShupe: Nope, just a query string (weird)
https://discordapp.com/developers/docs/resources/webhook#execute-slackcompatible-webhook-querystring-params

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
CoreyShupe: gotcha, time to Remodel then

src/main/java/com/github/princesslana/eriscasper/util/QueryStringBuilder.java
```diff
@@ -26,6 +26,10 @@ public QueryStringBuilder addSnowflake(String name, Optional<Snowflake> snowflak
     return snowflake.map(s -> add(name, s.unwrap())).orElse(this);
   }
 
+  public QueryStringBuilder addBoolean(String name, Optional<Boolean> value) {
+    return value.map(v -> add(name, v.toString())).orElse(this);
```
ianagbip1oti, line 30: Are we sure this outputs what we need? i.e., `true` vs `True` or something else?

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
```diff
@@ -0,0 +1,25 @@
...
+
+  BufferedImage image;
+
+  @Override
+  public String toString() {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    try {
+      ImageIO.write(image, "jpg", baos);
+    } catch (IOException e) {
+      e.printStackTrace();
```
ianagbip1oti: Don't output to stdout. Does it make sense to return empty string rather than throw an exception here?

src/main/java/com/github/princesslana/eriscasper/util/QueryStringBuilder.java
CoreyShupe: https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html I'm sure, needs true/false

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
```diff
@@ -0,0 +1,25 @@
+package com.github.princesslana.eriscasper.util;
+
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Base64;
+import javax.imageio.ImageIO;
+
+public abstract class SendableImage {
```
ianagbip1oti: We make a base class and just extend it in one place? Why not just one class?

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
CoreyShupe: we should probably throw an exception, it doesn't make sense to have an invalid image anywhere at all, especially sending it to discord.

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
CoreyShupe: It's extended in 2 places iirc AvatarImage and EmojiImage

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
```diff
@@ -0,0 +1,61 @@
...
+
+  AuditLogEventType(int code) {
+    this.code = code;
+  }
+
+  public int getCode() {
+    return code;
+  }
+
+  private static final AuditLogEventType[] types = new AuditLogEventType[72];
```
ianagbip1oti: This looks like premature optimization to me.

CoreyShupe: It's for end users to easily be able to find out what code corresponds to the audit log they're trying to look at. It's done the same way with OP CODE

ianagbip1oti: Finding an opcode from integer is fine. It's the storing of them in an array that I think is premature optimization.

Why not loop through all of them like OpCode does?

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
```diff
@@ -75,7 +77,7 @@ public String getPath() {
     return path;
   }
 
-  public Request.Builder newRequestBuilder(Rq rq) throws Exception {
+  Request.Builder newRequestBuilder(Rq rq) throws Exception {
```
ianagbip1oti: Why are we removing the public?

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
ianagbip1oti: Why do we choose jpg here?

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
CoreyShupe: got a warning saying it could be package-private, did it to everything which didn't need to be public, are we expecting end users to use this?

ianagbip1oti:
Can you provide an example of how someone makes use of the AvatarImage or EmojiImage class when sending a request? I think there might be something better than the class structure setup there.

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
CoreyShupe: AuditLogs tend to contain a very large sum of them, looping through them all will be slower than just grabbing the cached enum. It takes a very small amount of memory as well so why not prefer speed over memory.

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
CoreyShupe: jpq is more common than png, I'll look for a way to accept both but that's the way I found to move a image into Base64, I'll look int png and gif options in a few.

CoreyShupe:
```
@Value.Immutable
public interface CreateEmojiRequest {
  String getName();

  String getImage();

  ImmutableList<Snowflake> getRoles();
}
```
getImage needs a Base64 image, new EmojiImage(Image from file or wherever).toString() then you have it formatted
getData works the same for Avatar images, leaving the option to translate it to pure Base64 or translate it into data for changing the avatar image. Goes with some of the User commands.
```
@Value.Immutable
public interface ModifyUserRequest {
  String getUsername();

  String getAvatar();
}
```
getAvatar ^ should require formatted Avatar image from `getData` or just requires the Base64 data, haven't found any documentation for it except for the formatted data 
https://discordapp.com/developers/docs/resources/user#avatar-data
https://discordapp.com/developers/docs/resources/user#modify-current-user

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
ianagbip1oti: I prefer avoiding package visibility.

CoreyShupe: Alright, I'll move them to public rather than preferring package-private

ianagbip1oti:
So, I'm not a fan of using `toString` to produce the base64 of the image. `toString` is usually where you want to have info useful in logs/debugging. Not something essential to functionality.

But, I think it might be better to have the image passed into a constructor of the requests that use them. As it is now, the user has to know whether to call `toString` or `getData`, and that I need to use `EmojiImage` with the emoji request. I think we can come up with an API for the user to use that's a little more obvious and less open to errors.

I think `SendableImage#toString` can just be a static method somewhere. `SendableImage` is just being used to share code, we never make use of the `SendableImage` class anywhere.

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
ianagbip1oti: If we're going to be doing this array, then we need the `of` method to have some changes.

It shouldn't hard code `72`, we can use `types.length`, we should handle negative ints, and we should handle the case where an int is passed in that we don't have an entry for. To be consistent with OpCode an invalid value should result in an Exception.

CoreyShupe:
Fair enough, I'll change some implementation to compensate.

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
ianagbip1oti: jpg doesn't support transparency. Only gif supports animation. Forcing to a single image type isn't great behavior here.

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
CoreyShupe: Gotcha, will fix

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
CoreyShupe: Yeah I'm adding in implementation for other format types.

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
```diff
@@ -31,7 +31,9 @@
     GET("GET", Content.QUERY_STRING),
     PATCH("PATCH", Content.BODY),
     POST("POST", Content.BODY),
-    PUT("PUT", Content.BODY);
+    QUERY_POST("POST", Content.QUERY_STRING),
+    PUT("PUT", Content.BODY),
+    QUERY_PUT("PUT", Content.QUERY_STRING);
```
ianagbip1oti: Sending a request body along with these requests has not been addressed?

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
```diff
@@ -0,0 +1,19 @@
+package com.github.princesslana.eriscasper.util;
+
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayOutputStream;
+import java.util.Base64;
+
+public interface SendableImage {
```
ianagbip1oti: if we're just having utility methods, the style I've gone for is class with private constructor.

src/main/java/com/github/princesslana/eriscasper/util/FormatType.java
```diff
@@ -0,0 +1,32 @@
...
+import com.github.princesslana.eriscasper.ErisCasperFatalException;
+import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.io.OutputStream;
+import javax.imageio.ImageIO;
+
+public enum FormatType {
+  GIF("gif"),
+  JPG("jpg"),
+  JPEG("jpeg"),
```
ianagbip1oti: I don't think we need JPG and JPEG.

src/main/java/com/github/princesslana/eriscasper/rest/Route.java
CoreyShupe: yes it has, it doesn't need body content, check comments on your previous comment

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
CoreyShupe: gotcha, I'll change it

src/main/java/com/github/princesslana/eriscasper/util/FormatType.java
CoreyShupe: alright, they are different formats according to ImageIO which is why I added it

src/main/java/com/github/princesslana/eriscasper/rest/WebhookRoute.java
```diff
@@ -0,0 +1,142 @@
...
+   */
+  public Route<Void, Void> deleteWebhookWithToken(String token) {
+    return Route.delete(path("/%s", token), Void.class);
+  }
+
+  /**
+   * @see <a href="https://discordapp.com/developers/docs/resources/webhook#execute-webhook">
+   *     https://discordapp.com/developers/docs/resources/webhook#execute-webhook</a>
+   */
+  // TODO
```
ianagbip1oti: Let's not leave these big todo blogs in here. We can track this in #130

CoreyShupe: gotcha

src/main/java/com/github/princesslana/eriscasper/rest/webhook/ExecuteGithubWebhookRequest.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Optional;
+import org.immutables.value.Value;
+
+/**
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/webhook#execute-githubcompatible-webhook-querystring-params">
+ *     https://discordapp.com/developers/docs/resources/webhook#execute-githubcompatible-webhook-querystring-params</a>
+ */
+@Value.Immutable
+public interface ExecuteGithubWebhookRequest { // TODO unfinished see WebhookRoute
```
ianagbip1oti: Same as todos, let's not put unused code in the repo

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
```diff
@@ -0,0 +1,64 @@
...
+      types[type.getCode() - 1] = type;
+    }
+  }
+
+  public static AuditLogEventType of(int code) {
+    if (code > 72 || code < 0) {
+      throw new ErisCasperFatalException(
+          "Illegal code `" + code + "` found. Please refer to a code from 0-" + types.length + ".");
+    }
+    return types[code - 1];
```
ianagbip1oti: What about unknown codes? e.g., 55

CoreyShupe: ah I'll implement support for that

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
```diff
@@ -0,0 +1,64 @@
...
+
+  static {
+    for (AuditLogEventType type : AuditLogEventType.values()) {
+      types[type.getCode() - 1] = type;
+    }
+  }
+
+  public static AuditLogEventType of(int code) {
+    if (code > 72 || code < 0) {
+      throw new ErisCasperFatalException(
```
ianagbip1oti: ErisCasperFatalException isn't appropriate here.

ErisCasperFatalException should be used for cases where ErisCasper can not continue (yeah, that should be in the javadoc for the exception)

src/main/java/com/github/princesslana/eriscasper/rest/WebhookRoute.java
```diff
@@ -0,0 +1,109 @@
...
+   */
+  public Route<Void, Void> deleteWebhookWithToken(String token) {
+    return Route.delete(path("/%s", token), Void.class);
+  }
+
+  private String path(String fmt, String... args) {
+    return "/webhooks/" + id.unwrap() + String.format(fmt, Arrays.asList(args).toArray());
+  }
+
+  public WebhookRoute on(Snowflake webhookId) {
```
ianagbip1oti: should be static?

ianagbip1oti (approved):

PR #159, opened 2018-06-08
Update Route to allow multiple content in request

ianagbip1oti:


CoreyShupe (approved):
Looks good, probably better than the Tuple along with the Content enum.

Issue #123, opened 2018-03-29
Support Audit Log Resources


ianagbip1oti:
https://discordapp.com/developers/docs/resources/audit-log#audit-logs-resource

- [x] Audit Logs
- [x] Get Guild Audit Log

CoreyShupe:
Audit logs has no proper gateway for retrieval, what should be done with that there?

PR #160, opened 2018-06-09
AuditLog Resource addition

CoreyShupe:
Addresses https://github.com/princesslana/ErisCasper.java/issues/123


src/main/java/com/github/princesslana/eriscasper/rest/AuditLogRoute.java
```diff
@@ -0,0 +1,46 @@
...
+  // https://discordapp.com/developers/docs/resources/audit-log#get-guild-audit-log
+
+  private final Snowflake id;
+
+  private AuditLogRoute(Snowflake id) {
+    this.id = id;
+  }
+
+  /**
+   * Warning, this can and may throw an error. If it does please paste your console log into an
```
ianagbip1oti: Well, this is true of anything isn't it?

src/main/java/com/github/princesslana/eriscasper/rest/auditlog/GetGuildAuditLog.java
```diff
@@ -0,0 +1,37 @@
...
+import org.immutables.value.Value;
+
+/**
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/audit-log#get-guild-audit-log-query-string-parameters">
+ *     https://discordapp.com/developers/docs/resources/audit-log#get-guild-audit-log-query-string-parameters</a>
+ */
+@Value.Immutable
+public interface GetGuildAuditLog {
+  /*
```
ianagbip1oti: This just duplicates what is at the linked url in the javadoc?

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
```diff
@@ -0,0 +1,64 @@
...
+
+  static {
+    for (AuditLogEventType type : AuditLogEventType.values()) {
+      types[type.getCode() - 1] = type;
+    }
+  }
+
+  public static AuditLogEventType of(int code) {
+    AuditLogEventType type;
+    if (code > 72 || code < 1 || (type = types[code - 1]) == null) {
```
ianagbip1oti: Don't like the assignment in the if statement.

Why do we use 72 instead of types.length?

ianagbip1oti (changes requested):

src/main/java/com/github/princesslana/eriscasper/rest/AuditLogRoute.java
CoreyShupe: Especially true of this. Discord hasn't documented this well and I've run into undocumented incidents for this specific resource. It's more error prone so I felt it necessary to leave a comment about it.

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
CoreyShupe: Yeah I need to fix the if statement, and I'll change it, forgot about that piece.

CoreyShupe: The if statement is there to keep it in one if statement rather than 2.

src/main/java/com/github/princesslana/eriscasper/rest/AuditLogRoute.java
ianagbip1oti: Sorry, not a fan of this comment. More so the thing about filing an issue report (in areas where this comment isn't, I shouldn't file an issue report?).

An improvement would be something like 'These routes are not documented well in the Discord API documentation'

CoreyShupe: Alright, I'll change the commenting, didn't really know how to word it as to avoid what you pointed out in the first statement. I'll reword it.

src/main/java/com/github/princesslana/eriscasper/rest/auditlog/GetGuildAuditLog.java
CoreyShupe: yeap, noticed it late, removed

src/main/java/com/github/princesslana/eriscasper/rest/auditlog/GetGuildAuditLog.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Optional;
+import org.immutables.value.Value;
+
+/**
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/audit-log#get-guild-audit-log-query-string-parameters">
+ *     https://discordapp.com/developers/docs/resources/audit-log#get-guild-audit-log-query-string-parameters</a>
+ */
+@Value.Immutable
+public interface GetGuildAuditLog {
```
ianagbip1oti: Haven't we been adding a Request suffix to these?

src/main/java/com/github/princesslana/eriscasper/util/AuditLogEventType.java
```diff
@@ -0,0 +1,68 @@
...
+      types[type.getCode() - 1] = type;
+    }
+  }
+
+  public static AuditLogEventType of(int code) {
+    if (code > types.length || code < 1) {
+      throw new IllegalArgumentException(
+          "Illegal code `" + code + "` found. Please refer to a code from 1-" + types.length + ".");
+    }
+    AuditLogEventType type;
```
ianagbip1oti: This declaration doesn't have to be on it's own line

CoreyShupe: I don't have an explanation for this, idk why I did this. I'm probably just really tired right now.

ianagbip1oti (dismissed):

src/main/java/com/github/princesslana/eriscasper/rest/AuditLogRoute.java
```diff
@@ -0,0 +1,37 @@
+package com.github.princesslana.eriscasper.rest;
+
+import com.github.princesslana.eriscasper.data.Snowflake;
+import com.github.princesslana.eriscasper.data.resource.AuditLogObject;
+import com.github.princesslana.eriscasper.data.resource.Guild;
+import com.github.princesslana.eriscasper.rest.auditlog.GetGuildAuditLogRequest;
+
+public class AuditLogRoute {
+  // https://discordapp.com/developers/docs/resources/audit-log#get-guild-audit-log
```
ianagbip1oti: Should this be here?

CoreyShupe: Doesn't have to be, I'll remove it. I think I added it in for myself since I was coming back to it just due to the parsing complexity.

ianagbip1oti (approved):

Issue #125, opened 2018-03-29
Support Emoji Resource


ianagbip1oti:
https://discordapp.com/developers/docs/resources/emoji#emoji-resource

- [x] List Guild Emojis
- [x] Get Guild Emoji
- [x] Create Guild Emoji
- [x] Modify Guild Emoji
- [x] Delete Guild Emoji

PR #161, opened 2018-06-09
Emoji resource support.

CoreyShupe:
Addresses https://github.com/princesslana/ErisCasper.java/issues/125


src/main/java/com/github/princesslana/eriscasper/rest/EmojiRoute.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private EmojiRoute(Snowflake id) {
+    this.id = id;
+  }
+
+  /**
+   * @see <a href="https://discordapp.com/developers/docs/resources/emoji#list-guild-emojis">
+   *     https://discordapp.com/developers/docs/resources/emoji#list-guild-emojis</a>
+   */
+  public Route<Void, ImmutableList<Emoji>> getEmojis() {
```
ianagbip1oti: I'd like to follow the names in the discord docs, so this should be listGuildEmojis.

By being consistent with the discord docs, we have to write less docs ourselves, because it's easier to find things in the discord docs.

src/main/java/com/github/princesslana/eriscasper/rest/emoji/CreateEmojiRequest.java
```diff
@@ -0,0 +1,20 @@
...
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/emoji#create-guild-emoji-json-params">
+ *     https://discordapp.com/developers/docs/resources/emoji#create-guild-emoji-json-params</a>
+ */
+@Value.Style(deepImmutablesDetection = true)
+@Value.Immutable
+public interface CreateEmojiRequest {
+  String getName();
+
+  EmojiImageWrapper getImage();
```
ianagbip1oti: I'm undecided on this. I think my preferred way is to have this be String, and provide convenience methods that allow a File or whatever to be passed in.

src/main/java/com/github/princesslana/eriscasper/util/FormatType.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.io.OutputStream;
+import javax.imageio.ImageIO;
+
+public enum FormatType {
+  GIF("gif"),
+  JPG("jpg"),
+  JPEG("jpeg"),
```
ianagbip1oti: So, you said ImageIO supports both. What is the difference between them?

ianagbip1oti (changes requested):

src/main/java/com/github/princesslana/eriscasper/rest/EmojiRoute.java
CoreyShupe: alright

src/main/java/com/github/princesslana/eriscasper/rest/emoji/CreateEmojiRequest.java
CoreyShupe: you can use a string in the current setup

src/main/java/com/github/princesslana/eriscasper/util/FormatType.java
CoreyShupe: not sure lemme look into it

src/main/java/com/github/princesslana/eriscasper/util/SendableImage.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Base64;
+
+public final class SendableImage {
+
+  private SendableImage() {}
+
+  public static String imageToBase64(File file) throws IOException {
+    InputStream stream = file.toURI().toURL().openConnection().getInputStream();
```
ianagbip1oti: This stream is never closed?

Look into Guava's IO stuff: https://github.com/google/guava/wiki/IOExplained. I think you'll want to start with `Files.asByteSource`

ianagbip1oti (approved):

Issue #127, opened 2018-03-29
Support Invite Resource


ianagbip1oti:
https://discordapp.com/developers/docs/resources/invite#invite-resource

- [x] Get Invite
- [x] Delete Invite

We will not support 'Accept Invite' as:
> This endpoint is deprecated and will be discontinued on March 23, 2018. Add Guild Member should be used in its place.

PR #162, opened 2018-06-10
Invite resources.

CoreyShupe:
Addresses https://github.com/princesslana/ErisCasper.java/issues/127

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/rest/InviteRoute.java
+++ b/src/main/java/com/github/princesslana/eriscasper/rest/InviteRoute.java
@@ -1,0 +1,38 @@
+package com.github.princesslana.eriscasper.rest;
+
+import com.github.princesslana.eriscasper.data.resource.Invite;
+import java.util.Optional;
+
+public class InviteRoute {
+
+  private final String path;
+
+  private InviteRoute(String code) {
+    this.path = "/invites/" + code;
+  }
+
+  /**
+   * @see <a href="https://discordapp.com/developers/docs/resources/invite#get-invite">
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/princesslana/eriscasper/rest/InviteRoute.java
```diff
@@ -0,0 +1,38 @@
...
+    this.path = "/invites/" + code;
+  }
+
+  /**
+   * @see <a href="https://discordapp.com/developers/docs/resources/invite#get-invite">
+   *     https://discordapp.com/developers/docs/resources/invite#get-invite</a>
+   */
+  public Route<Void, Invite> get(Optional<Boolean> withCounts) {
+    return Route.get(
+        path + withCounts.map(bool -> ";with_counts=" + bool).orElse(""), Invite.class);
```
ianagbip1oti: Is this mean to be a ? instead of ;

ianagbip1oti: Why not use QueryStringBuilder?

CoreyShupe: Nope
https://stackoverflow.com/questions/39266970/what-is-the-difference-between-url-parameters-and-query-strings
https://discordapp.com/developers/docs/resources/invite#get-invite

CoreyShupe: It's not a QueryString

ianagbip1oti (approved):

Issue #126, opened 2018-03-29
Support Guild Resource


ianagbip1oti:
https://discordapp.com/developers/docs/resources/guild#guild-resource

- [x] Create Guild
- [x] Get Guild
- [x] Modify Guild
- [x] Delete Guild
- [x] Get Guild Channels
- [x] Create Guild Channel
- [x] Modify Guild Channel Positions
- [x] Get Guild Member
- [x] List Guild Members
- [x] Add Guild Member
- [x] Modify Guild Member
- [x] Modify Current User Nick
- [x] Add Guild Member Role
- [x] Remove Guild Member Role
- [x] Remove Guild Member
- [x] Get Guild Bans
- [x] Create Guild Ban
- [x] Remove Guild Ban
- [x] Get Guild Roles
- [x] Create Guild Role
- [x] Modify Guild Role Positions
- [x] Modify Guild Role
- [x] Delete Guild Role
- [x] Get Guild Prune Count
- [x] Begin Guild Prune
- [x] Get Guild Voice Regions
- [x] Get Guild Invites
- [x] Get Guild Integrations
- [x] Create Guild Integration
- [x] Modify Guild Integration
- [x] Delete Guild Integration
- [x] Sync Guild Integration
- [x] Get Guild Embed
- [x] Modify Guild Embed
- [x] Get Guild Vanity URL

PR #163, opened 2018-06-10
Guild Resource Support

CoreyShupe:
Addresses https://github.com/princesslana/ErisCasper.java/issues/126

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
+++ b/src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
@@ -176,13 +176,8 @@ public class GuildRoute {
    * @see <a href="https://discordapp.com/developers/docs/resources/guild#get-guild-ban">
    *     https://discordapp.com/developers/docs/resources/guild#get-guild-ban</a>
    */
-  public Route<Void, Optional<Ban>> getGuildBan(Snowflake userId) {
-    return Route.get(
-        path("/bans/%s" + userId.unwrap()),
-        (response) ->
-            response.code() == 404
-                ? Optional.empty()
-                : Optional.of(Data.fromJson(response.body().string(), Ban.class)));
+  public Route<Void, Ban> getGuildBan(Snowflake userId) {
+    return Route.get(path("/bans/%s" + userId.unwrap()), Ban.class);
   }
 
   /**
@@ -350,13 +345,8 @@ public class GuildRoute {
    * @see <a href="https://discordapp.com/developers/docs/resources/guild#get-guild-vanity-url">
    *     https://discordapp.com/developers/docs/resources/guild#get-guild-vanity-url</a>
    */
-  public Route<Void, Optional<PartialInvite>> getGuildVanityUrl() {
-    return Route.get(
-        path("/vanity-url"),
-        response ->
-            response.code() == 204
-                ? Optional.empty()
-                : Optional.of(Data.fromJson(response.body().string(), PartialInvite.class)));
+  public Route<Void, PartialInvite> getGuildVanityUrl() {
+    return Route.get(path("/vanity-url"), PartialInvite.class);
   }
 
   private String path(String fmt, String... args) {
@@ -375,7 +365,7 @@ public class GuildRoute {
    * @see <a href="https://discordapp.com/developers/docs/resources/guild#create-guild">
    *     https://discordapp.com/developers/docs/resources/guild#create-guild</a>
    */
-  public static Route<GuildCreateRequest, Guild> createNewGuild() {
+  public static Route<GuildCreateRequest, Guild> createGuild() {
     return Route.post("/guilds", GuildCreateRequest.class, Guild.class);
   }
 }
--- a/src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
+++ b/src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
@@ -21,8 +21,8 @@ import com.github.princesslana.eriscasper.rest.guild.GetGuildPruneCountRequest;
 import com.github.princesslana.eriscasper.rest.guild.GuildChannelCreateRequest;
 import com.github.princesslana.eriscasper.rest.guild.GuildCreateRequest;
 import com.github.princesslana.eriscasper.rest.guild.ListGuildMemberRequest;
-import com.github.princesslana.eriscasper.rest.guild.ModifyGuildChannelPositionsRequest;
 import com.github.princesslana.eriscasper.rest.guild.ModifyCurrentNickRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyGuildChannelPositionsRequest;
 import com.github.princesslana.eriscasper.rest.guild.ModifyGuildIntegrationRequest;
 import com.github.princesslana.eriscasper.rest.guild.ModifyGuildMemberRequest;
 import com.github.princesslana.eriscasper.rest.guild.ModifyGuildRequest;
@@ -305,7 +305,8 @@ public class GuildRoute {
    * @see <a href="https://discordapp.com/developers/docs/resources/guild#modify-guild-integration">
    *     https://discordapp.com/developers/docs/resources/guild#modify-guild-integration</a>
    */
-  public Route<ModifyGuildIntegrationRequest, Void> modifyGuildIntegration(Snowflake integrationId) {
+  public Route<ModifyGuildIntegrationRequest, Void> modifyGuildIntegration(
+      Snowflake integrationId) {
     return Route.patch(
         path("/integrations/%s", integrationId.unwrap()),
         ModifyGuildIntegrationRequest.class,
--- a/src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
+++ b/src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
@@ -1,8 +1,37 @@
 package com.github.princesslana.eriscasper.rest;
 
+import com.github.princesslana.eriscasper.data.Data;
 import com.github.princesslana.eriscasper.data.Snowflake;
+import com.github.princesslana.eriscasper.data.resource.Ban;
+import com.github.princesslana.eriscasper.data.resource.Channel;
 import com.github.princesslana.eriscasper.data.resource.Guild;
+import com.github.princesslana.eriscasper.data.resource.GuildEmbed;
+import com.github.princesslana.eriscasper.data.resource.GuildMember;
+import com.github.princesslana.eriscasper.data.resource.Integration;
+import com.github.princesslana.eriscasper.data.resource.InviteWithMetadata;
+import com.github.princesslana.eriscasper.data.resource.PartialInvite;
+import com.github.princesslana.eriscasper.data.resource.Role;
+import com.github.princesslana.eriscasper.data.resource.VoiceRegion;
+import com.github.princesslana.eriscasper.rest.guild.AddGuildMemberRequest;
+import com.github.princesslana.eriscasper.rest.guild.BeginGuildPruneRequest;
+import com.github.princesslana.eriscasper.rest.guild.CreateGuildBanRequest;
+import com.github.princesslana.eriscasper.rest.guild.CreateGuildIntegrationRequest;
+import com.github.princesslana.eriscasper.rest.guild.CreateGuildRoleRequest;
+import com.github.princesslana.eriscasper.rest.guild.GetGuildPruneCountRequest;
+import com.github.princesslana.eriscasper.rest.guild.GuildChannelCreateRequest;
+import com.github.princesslana.eriscasper.rest.guild.GuildCreateRequest;
+import com.github.princesslana.eriscasper.rest.guild.ListGuildMemberRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyGuildChannelPositionsRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyCurrentNickRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyGuildIntegrationRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyGuildMemberRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyGuildRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyGuildRolePositionsRequest;
+import com.github.princesslana.eriscasper.rest.guild.ModifyGuildRoleRequest;
+import com.github.princesslana.eriscasper.util.Pruned;
+import com.google.common.collect.ImmutableList;
 import java.util.Arrays;
+import java.util.Optional;
 
 public class GuildRoute {
   private final Snowflake id;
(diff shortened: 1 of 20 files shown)


src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
```diff
@@ -11,22 +40,323 @@ private GuildRoute(Snowflake id) {
...
+
+  /**
+   * @see <a href="https://discordapp.com/developers/docs/resources/guild#get-guild-ban">
+   *     https://discordapp.com/developers/docs/resources/guild#get-guild-ban</a>
+   */
+  public Route<Void, Optional<Ban>> getGuildBan(Snowflake userId) {
+    return Route.get(
+        path("/bans/%s" + userId.unwrap()),
+        (response) ->
+            response.code() == 404
```
ianagbip1oti: We shouldn't be dealing with 404 here. Most likely this throws an exception when the response is received anyway.

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
```diff
@@ -11,22 +40,323 @@ private GuildRoute(Snowflake id) {
...
+
+  /**
+   * @see <a href="https://discordapp.com/developers/docs/resources/guild#get-guild-vanity-url">
+   *     https://discordapp.com/developers/docs/resources/guild#get-guild-vanity-url</a>
+   */
+  public Route<Void, Optional<PartialInvite>> getGuildVanityUrl() {
+    return Route.get(
+        path("/vanity-url"),
+        response ->
+            response.code() == 204
```
ianagbip1oti: I don't see this 204 code in the docs. This has been tested?

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
CoreyShupe: No, it should be a clean 404 according to documentation.

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
```diff
@@ -40,4 +370,12 @@ public static GuildRoute on(Snowflake id) {
   public static GuildRoute on(Guild guild) {
     return on(guild.getId());
   }
+
+  /**
+   * @see <a href="https://discordapp.com/developers/docs/resources/guild#create-guild">
+   *     https://discordapp.com/developers/docs/resources/guild#create-guild</a>
+   */
+  public static Route<GuildCreateRequest, Guild> createNewGuild() {
```
ianagbip1oti: s/createNewGuild/createGuild/ (match discord name)

src/main/java/com/github/princesslana/eriscasper/rest/guild/AddGuildMemberRequest.java
```diff
@@ -0,0 +1,23 @@
...
+import com.google.common.collect.ImmutableList;
+import org.immutables.value.Value;
+
+/**
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/guild#add-guild-member-json-params">
+ *     https://discordapp.com/developers/docs/resources/guild#add-guild-member-json-params</a>
+ */
+@Value.Immutable
+public interface AddGuildMemberRequest {
```
ianagbip1oti, line 14: All these fields should be optional?

CoreyShupe: didn't see that, will fix

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
CoreyShupe: will change

src/main/java/com/github/princesslana/eriscasper/rest/guild/BeginGuildPruneRequest.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+@Value.Immutable
+public interface BeginGuildPruneRequest {
+  Integer getDays();
+
+  default String toQueryString() {
+    return new QueryStringBuilder().add("days", getDays().toString()).build();
+  }
+
+  static BeginGuildPruneRequest from(int days) {
```
ianagbip1oti: I believe `of` would be a more consistent name used by immutables.

Do we have these methods in these classes elsewhere?

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
CoreyShupe: I forget why I did that. I'm sure I saw something, but we can change it to a clean change into PartialInvite.

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
ianagbip1oti: Anything other than a 2xx response is treated as an error at the time we get the http response.

Handling this in a way other than an exception is a job for layers of code above this.

src/main/java/com/github/princesslana/eriscasper/rest/guild/BeginGuildPruneRequest.java
CoreyShupe: in a few of them which only require one param or simple calls, similar to what was done with CreateMessage.

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
CoreyShupe: We should handle it above then because we are expecting a 404 when there's no ban set for a user.

src/main/java/com/github/princesslana/eriscasper/rest/guild/BeginGuildPruneRequest.java
ianagbip1oti: Cool. I thought we had put those methods elsewhere. `of` is the more consistent name.

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildBanRequest.java
```diff
@@ -0,0 +1,23 @@
...
+@Value.Immutable
+public interface CreateGuildBanRequest {
+  Integer getDeleteMessageDays();
+
+  String getReason();
+
+  default String toQueryString() {
+    return new QueryStringBuilder()
+        .add("reason", getReason())
+        .add("delete_message_days", getDeleteMessageDays().toString())
```
ianagbip1oti: Shouldn't this make use of addLong?

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildBanRequest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.immutables.value.Value;
+
+/**
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/guild#create-guild-ban-query-string-params">
+ *     https://discordapp.com/developers/docs/resources/guild#create-guild-ban-query-string-params</a>
+ */
+@Value.Immutable
+public interface CreateGuildBanRequest {
+  Integer getDeleteMessageDays();
```
ianagbip1oti: We've used Long for integer fields elsewhere

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildBanRequest.java
CoreyShupe: it's non optional

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildBanRequest.java
CoreyShupe: yeah I changed it

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
ianagbip1oti: Sure, we're expecting 404 in a number of places I imagine (e.g., if they try an get a guild with an id that doesn't exist)

src/main/java/com/github/princesslana/eriscasper/rest/guild/BeginGuildPruneRequest.java
CoreyShupe: Gotcha, I'll change it.

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildBanRequest.java
ianagbip1oti: You could call with Optional.of(getDeleteMessageDays()) or add an `addLong(String, Long)` method to QueryStringBuilder.

CoreyShupe: Alright, I'll do that.

src/main/java/com/github/princesslana/eriscasper/rest/guild/ListGuildMemberRequest.java
```diff
@@ -0,0 +1,42 @@
...
+
+  static ListGuildMemberRequest withLimit(Long limit) {
+    return withLimitAfter(limit, null);
+  }
+
+  static ListGuildMemberRequest after(Snowflake id) {
+    return withLimitAfter(null, id);
+  }
+
+  static ListGuildMemberRequest withLimitAfter(@Nullable Long limit, @Nullable Snowflake after) {
```
ianagbip1oti: By the time we get to adding methods for all possible combos I think we gain not much. It would also be nice to avoid nullable fields in public methods.

src/main/java/com/github/princesslana/eriscasper/rest/guild/ModifyCurrentNickRequest.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @see <a
+ *     href="https://discordapp.com/developers/docs/resources/guild#modify-current-user-nick-json-params">
+ *     https://discordapp.com/developers/docs/resources/guild#modify-current-user-nick-json-params</a>
+ */
+@Value.Immutable
+public interface ModifyCurrentNickRequest {
+  String getNick();
+
+  static ModifyCurrentNickRequest withNick(String nick) {
```
ianagbip1oti: `ofNick`.

It does sound a little funky, but we use of for construction of classes elsewhere. `withXxx` methods are generated by immutables and have a different use.

src/main/java/com/github/princesslana/eriscasper/rest/guild/ListGuildMemberRequest.java
CoreyShupe: Alright, I'll remove them then, you're right.

src/main/java/com/github/princesslana/eriscasper/rest/guild/ModifyCurrentNickRequest.java
CoreyShupe: gotcha

src/main/java/com/github/princesslana/eriscasper/rest/GuildRoute.java
CoreyShupe: Alright, makes sense, they are able to get all the bans, so it might not be important to watch for it here.

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildRoleRequest.java
```diff
@@ -0,0 +1,30 @@
...
+
+  Optional<Integer> getPermissions();
+
+  Optional<Integer> getColor();
+
+  Optional<Boolean> isHoist();
+
+  Optional<Boolean> isMentionable();
+
+  static CreateGuildRoleRequest newRole() {
```
ianagbip1oti: Would would we expect this to be used for?

src/main/java/com/github/princesslana/eriscasper/rest/guild/GetGuildPruneCountRequest.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+@Value.Immutable
+public interface GetGuildPruneCountRequest {
+  Integer getDays();
+
+  default String toQueryString() {
+    return new QueryStringBuilder().add("days", getDays().toString()).build();
+  }
+
+  static GetGuildPruneCountRequest from(int days) {
```
ianagbip1oti: s/from/of

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildRoleRequest.java
```diff
@@ -0,0 +1,30 @@
...
+    return ImmutableCreateGuildRoleRequest.builder().build();
+  }
+
+  static CreateGuildRoleRequest of(String name) {
```
ianagbip1oti: Is naming this `ofName` more consistent with other `of` methods we have?

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildRoleRequest.java
CoreyShupe: Yeah, I mean just making a standard empty new role to work with later, works the same way in the client, just creates a default role to work with and position. If you want I can remove it.

ianagbip1oti: Do we think people would create a new role and update it with a bot? Or just create the role with the details they need in the create request?

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildRoleRequest.java
CoreyShupe: Not sure, we kind of switch, `of` not `ofDays` `ofNick` not `of`, so I mean it kind of goes back and forth.

src/main/java/com/github/princesslana/eriscasper/rest/guild/GetGuildPruneCountRequest.java
CoreyShupe: Gotcha

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildRoleRequest.java
CoreyShupe: We should decide on a format and be consistent with it, it's mostly my fault for switching in between, just didn't know of the right format to use.

ianagbip1oti: I thought we might.

For me I think we should stick with `ofXxx` going forward, because using just `of` breaks down if there's multiple of's I might want that take a String.

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildRoleRequest.java
CoreyShupe: You're right, I'll remove it. Most likely situation is creating with information.

src/main/java/com/github/princesslana/eriscasper/rest/guild/CreateGuildRoleRequest.java
CoreyShupe: Alright, I agree, should I change all past `of`s, or just from now on be consistent with ofXxx.

ianagbip1oti: Nah, we don't need to change past ones yet. Let's be consistent going forward, and if we're woking on a class or functionality that has an `of`, we can change it then.

CoreyShupe: Alright, sounds good, I'll change all of GuildRoute's `of`s just to keep this PR consistent and think about it going forward.

src/main/java/com/github/princesslana/eriscasper/rest/guild/GuildChannelCreateRequest.java
```diff
@@ -0,0 +1,39 @@
...
+
+  Optional<Snowflake> getParentId();
+
+  Optional<Boolean> isNsfw();
+
+  static GuildChannelCreateRequest ofName(String name) {
+    return ImmutableGuildChannelCreateRequest.builder().name(name).build();
+  }
+
+  static GuildChannelCreateRequest ofNameAndCategory(String name, Snowflake category) {
```
ianagbip1oti, line 36: This is where of will start to be troublesome, cos methods names with And in them are usually bad.

But we'll keep this for now and see how it goes.

ianagbip1oti (approved):

src/main/java/com/github/princesslana/eriscasper/rest/guild/GuildChannelCreateRequest.java
CoreyShupe: what makes them bad?

PR #165, opened 2018-06-11
Guild/Channel repository updates.

CoreyShupe:
- Separated guild / channel repos.
- Added more events to each watching for most updates to items.
- Added more persistence stuff.
- Added tests to test every event.

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
+++ b/src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
@@ -2,13 +2,8 @@ package com.github.princesslana.eriscasper.repository;
 
 import com.github.princesslana.eriscasper.data.Snowflake;
 import com.github.princesslana.eriscasper.data.resource.Channel;
-import com.google.common.collect.ImmutableMap;
 import io.reactivex.Maybe;
 import io.reactivex.Observable;
-import io.reactivex.Single;
-import io.reactivex.annotations.NonNull;
-import io.reactivex.annotations.Nullable;
-import io.reactivex.functions.Predicate;
 
 public interface ChannelRepository {
   Maybe<Channel> getChannel(Snowflake id);
--- a/src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
+++ b/src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
@@ -11,20 +11,7 @@ import io.reactivex.annotations.Nullable;
 import io.reactivex.functions.Predicate;
 
 public interface ChannelRepository {
-  Maybe<Channel> getChannel(@NonNull Snowflake id);
+  Maybe<Channel> getChannel(Snowflake id);
 
-  Single<ImmutableMap<Snowflake, Channel>> getChannels();
-
-  Maybe<Channel> getGuildCategoryFromName(Snowflake guildId, String name);
-
-  Maybe<Channel> getGuildChannelFromName(
-      Snowflake guildId, @Nullable Snowflake category, String name);
-
-  Observable<Channel> getGuildCategories(Snowflake guildId);
-
-  Observable<Channel> getGuildChannelsInCategory(Snowflake guildId, @Nullable Snowflake category);
-
-  Observable<Channel> filter(Predicate<Channel> channelPredicate);
-
-  Observable<Channel> filterGuild(Snowflake guildId, Predicate<Channel> channelPredicate);
+  Observable<Channel> getChannels();
 }
--- a/src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
+++ b/src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
@@ -1,0 +1,30 @@
+package com.github.princesslana.eriscasper.repository;
+
+import com.github.princesslana.eriscasper.data.Snowflake;
+import com.github.princesslana.eriscasper.data.resource.Channel;
+import com.google.common.collect.ImmutableMap;
+import io.reactivex.Maybe;
+import io.reactivex.Observable;
+import io.reactivex.Single;
+import io.reactivex.annotations.NonNull;
+import io.reactivex.annotations.Nullable;
+import io.reactivex.functions.Predicate;
+
+public interface ChannelRepository {
+  Maybe<Channel> getChannel(@NonNull Snowflake id);
+
+  Single<ImmutableMap<Snowflake, Channel>> getChannels();
+
+  Maybe<Channel> getGuildCategoryFromName(Snowflake guildId, String name);
+
+  Maybe<Channel> getGuildChannelFromName(
+      Snowflake guildId, @Nullable Snowflake category, String name);
+
+  Observable<Channel> getGuildCategories(Snowflake guildId);
+
+  Observable<Channel> getGuildChannelsInCategory(Snowflake guildId, @Nullable Snowflake category);
+
+  Observable<Channel> filter(Predicate<Channel> channelPredicate);
+
+  Observable<Channel> filterGuild(Snowflake guildId, Predicate<Channel> channelPredicate);
+}
--- a/src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
+++ b/src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
@@ -20,11 +20,7 @@ import com.github.princesslana.eriscasper.rx.Maybes;
 import com.google.common.collect.ImmutableMap;
 import io.reactivex.Maybe;
 import io.reactivex.Observable;
-import io.reactivex.Single;
-import io.reactivex.annotations.NonNull;
-import io.reactivex.annotations.Nullable;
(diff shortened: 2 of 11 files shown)


src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
```diff
@@ -0,0 +1,30 @@
...
+import com.google.common.collect.ImmutableMap;
+import io.reactivex.Maybe;
+import io.reactivex.Observable;
+import io.reactivex.Single;
+import io.reactivex.annotations.NonNull;
+import io.reactivex.annotations.Nullable;
+import io.reactivex.functions.Predicate;
+
+public interface ChannelRepository {
+  Maybe<Channel> getChannel(@NonNull Snowflake id);
```
ianagbip1oti: No need for `@NonNull` here. By convention we assume non null

src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
```diff
@@ -0,0 +1,30 @@
...
+
+  Single<ImmutableMap<Snowflake, Channel>> getChannels();
```
ianagbip1oti: Why do we return a Map here, rather than something like `Observable<Channel>`?

src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
```diff
@@ -0,0 +1,30 @@
...
+import com.github.princesslana.eriscasper.data.resource.Channel;
+import com.google.common.collect.ImmutableMap;
+import io.reactivex.Maybe;
+import io.reactivex.Observable;
+import io.reactivex.Single;
+import io.reactivex.annotations.NonNull;
+import io.reactivex.annotations.Nullable;
+import io.reactivex.functions.Predicate;
+
+public interface ChannelRepository {
```
ianagbip1oti: I'm hesitant to have some many methods in the Repository.

The Repositories we want to have multiple implementations (e.g,. we might have one that caches in Redis, we might have one that uses a DB, we might have one that fetches info via the routes), so keeping it smaller makes it easier to implement these.

Those methods that are useful, but can be implemented purely by calling other methods of the repository I think we can move to the convenience layer.

src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
```diff
@@ -0,0 +1,207 @@
...
+            map.replace(channel.getId(), channel);
+            return ImmutableMap.copyOf(map);
+          });
+
+  private final ConnectableObservable<ImmutableMap<Snowflake, Channel>> channelWatcher;
+
+  public ChannelsFromEvents(Observable<Event> events) {
+    // Merge all events which are meant to modify the channel by any means
+    channelWatcher =
+        events
```
ianagbip1oti: This `events.ofType(clz).map(unwrap).map(doSeomthing)` pattern looks like a candidate for a method.

src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
```diff
@@ -0,0 +1,207 @@
...
+                    .ofType(ChannelPinsUpdateEvent.class)
+                    .map(ChannelPinsUpdateEvent::unwrap)
+                    .map(PINS_UPDATE_CHANNEL_FUNCTION))
+            .mergeWith(
+                events
+                    .ofType(ChannelUpdateEvent.class)
+                    .map(ChannelUpdateEvent::unwrap)
+                    .map(UPDATE_CHANNEL_FUNCTION))
+            .scan(ImmutableMap.<Snowflake, Channel>of(), (map, function) -> function.apply(map))
+            .doOnError(Throwable::printStackTrace)
```
ianagbip1oti: Don't use stdout

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -1,168 +1,275 @@
...
+  private static final GuildFunctionData<GuildRoleUpdateEventData> ROLE_UPDATE_GUILD_FUNCTION =
+      GuildFunctionData.of(
+          (map, data) -> {
+            Guild guild = map.get(data.getGuildId());
+            Snowflake id = data.getRole().getId();
+            List<Role> roles = new ArrayList<>(guild.getRoles());
+            Role initial =
+                roles.stream().filter(role -> role.getId().equals(id)).findAny().orElse(null);
+            if (initial == null) {
+              throw new RuntimeException("Data not found where it should be.");
```
ianagbip1oti: You could use `orElseThrow` instead of `orElse(null)`. I think we can do better than `RuntimeException`. Perhaps `IllegalStateException`?

src/test/java/com/github/princesslana/eriscasper/faker/DataFaker.java
```diff
@@ -78,6 +84,32 @@ public static UnavailableGuild unavailableGuildFromGuild(Snowflake predetermined
     return ImmutableUnavailableGuild.builder().id(predeterminedId).build();
   }
 
+  public static Emoji emoji() {
+    return ImmutableEmoji.builder().addRoles(DiscordFaker.snowflake()).name(username()).build();
```
ianagbip1oti: While `username` does what we want, it's not semantically clear, cos an emoji name is not a username

src/test/java/com/github/princesslana/eriscasper/faker/DataFaker.java
```diff
@@ -78,6 +84,32 @@ public static UnavailableGuild unavailableGuildFromGuild(Snowflake predetermined
...
+  }
+
+  public static GuildMember guildMember() {
```
ianagbip1oti: Can't we add more randomness to these? e.g., a random boolean or long, rather than something constant.

src/test/java/com/github/princesslana/eriscasper/repository/event/TestChannelsFromEvents.java
```diff
@@ -0,0 +1,177 @@
...
+  @BeforeMethod
+  public void subject() {
+    subject = new ChannelsFromEvents(events);
+  }
+
+  @Test
+  public void getChannel_whenGuildCreated_shouldCacheChannel() {
+    TestObserver<Channel> observer = new TestObserver<>();
+
+    // Create guild, define channel
```
ianagbip1oti: If we're going to leave these comments in the tests let's stick with the semi-standard given, when, then.

Some of these comments are very close to just duplicating what the line of code says rather than adding info

ianagbip1oti (changes requested):

src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
CoreyShupe: we could do that but for convenience, I might add both implementations just for variety and usability.

src/main/java/com/github/princesslana/eriscasper/repository/ChannelRepository.java
CoreyShupe: Gotcha, I was hesitant as well but I kinda figured some people don't know the number values of channel types, so I setup category lookups to serve that.

src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
CoreyShupe: Agreed.

src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
CoreyShupe: Yeah, I need to do that, my bad.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
CoreyShupe: IllegalStateException works, I want to handle the exception myself instead of letting the optional throw.

src/test/java/com/github/princesslana/eriscasper/faker/DataFaker.java
CoreyShupe: hmm okay, I'll look into more string fakers.

src/test/java/com/github/princesslana/eriscasper/repository/event/TestChannelsFromEvents.java
CoreyShupe: yeah I'm going to remove the comments, was just something for me to remember what I was doing. It's not necessary to have them at all.

src/test/java/com/github/princesslana/eriscasper/faker/DataFaker.java
CoreyShupe: I'm sure we can, I'll look into more randominess

ianagbip1oti (approved):

Issue #120, opened 2018-03-29
Stage 1 Sharding


ianagbip1oti:
https://discordapp.com/developers/docs/topics/gateway#sharding

Accept EC_SHARD_ID and EC_SHARD_NUM as environment variables and send these to Discord.

Part of #105

CoreyShupe:
I think sharding should be done though the create method:
ErisCasper.create(token, max_shards, shard_id)
I only say this since using them as environment variables stops it from being capable of running multiple shards in the same program, which could be useful for the user.

ianagbip1oti:
Sure, there would be a method like this, but the default `create` method should check the environment variables.

CoreyShupe:
I was thinking system property variables rather than environment variables since it makes more sense imo. Environment variables are supposed to be static throughout the machine, system properties are manipulated per program. So I just setup system properties to handle stage 1 sharding along with param values.

ianagbip1oti:
For using system properties as well as env vars see #106 

Using environment variables config I want to support as it's a best practice: https://12factor.net/config

CoreyShupe:
I implemented system properties as well as env variables for handling the bot token, but we don't currently have a config system setup? I could implement environment variables for sharding, it wouldn't take too too much time.

ianagbip1oti:
Yes, we don't currently have a config system setup, that's what #106 is about.

I would prefer using a consistent method of config (being environment variables) until #106 is in place and supports other methods

Issue #71, opened 2018-03-15
Handle bad token


ianagbip1oti:
Handle cleanly case when the bot token is bad.

Perhaps by throwing an ErisCasperFatalException?

WangzoR:
assuming that EC_TOKEN will be the permanent method of handling the bot token
worthwhile putting in an exception and check for null values returned as part of the get system.env on top of the primary check which is to gracefully handle if token is totally invalid

WangzoR:
@ianagbip1oti, some guidance please, so I made an attempt at the null check, but after I finished and started digging deeper it seems intrinsically built in already as part of the BotToken object, so not so sure at what 'level' this kind of exception check should be handled in

**My Code:**
![image](https://user-images.githubusercontent.com/37465280/38233959-b0db936e-375f-11e8-8018-cf2bbf52a9e2.png)
![image](https://user-images.githubusercontent.com/37465280/38233975-bf08e626-375f-11e8-85f4-2ac8d2cc448d.png)

It appears that this already throws a semi suitable exception, just change the message to suit?
**Existing lower level code:**
![image](https://user-images.githubusercontent.com/37465280/38234018-e1fab7ea-375f-11e8-8cc6-88209a82993a.png)

**Format validity?**
In addition, for token validity sake, aside from DiscordAPI throwing a fit at connecting, is there any baseline integrity format check? found the below, no idea what this represents and significance of 5381 and if this is to utilised to verify anything?
![image](https://user-images.githubusercontent.com/37465280/38234070-0f73af92-3760-11e8-926f-7f9d9699c08a.png)

ianagbip1oti:
`EC_TOKEN` will always be a way of passing in the token, it may not be the only way (see #106), but since #106 is not done yet, just handling `EC_TOKEN` here is fine.

I don't think we need to check the token format. If a token is provided we attempt to login with it, and handle the error message Discord sends us back.

Otherwise yeah, we just want more user friendly error messages in the exception.

There is no significance to the 5381 in `hashCode`. Choosing some kind of randomish seed for use in `hashCode` is fairly common.

CoreyShupe:
Handled bad token by catching authentication issues given by discord. Starts to go bad once you hit the rate limit but that should be worried about by the client. I do null checks for the EC_TOKEN however if the token is for example `test` it's allowed through but caught immediately by discord.

ianagbip1oti:
What rate limit are you referring to here?

[...]

Issue #122, opened 2018-03-29
Support all Gateway Commands


ianagbip1oti:
https://discordapp.com/developers/docs/topics/gateway#commands-and-events-gateway-commands

- [ ] Identify | triggers the initial handshake with the gateway
- [x] Resume | resumes a dropped gateway connection
- [ ] Heartbeat | maintains an active gateway connection
- [ ] Request Guild Members | requests offline members for a guild
- [ ] Update Voice State | joins, moves, or disconnects the client from a voice channel
- [ ] Update Status | updates a client's presence

ianagbip1oti:
A couple of things TBD:
* Should Commands have a wrapper like Events do?
* Heartbeat is supported, but should probably be broken out into a class for consistency? https://github.com/princesslana/ErisCasper.java/blob/master/src/main/java/com/github/princesslana/eriscasper/gateway/Payloads.java#L32-L34

CoreyShupe:
If you break out all the commands into separate classes they should be separate not inner classes (like resume and identify are currently inner classes which maybe should be broken out similar to how routes are setup) Also what would the command wrappers even look like? They don't really generate a response.

Although heartbeats shouldn't be separated imo since it's just a sequence number being passed, anything more would over complicate things. Especially since a heartbeat will never be sent by a user. Everything else being wrapped in a separate class has a use.

Issue #105, opened 2018-03-25
Sharding


ianagbip1oti:
https://discordapp.com/developers/docs/topics/gateway#sharding

Sharding can be implemented in multiple stages. I think Stage 1 is all that is required to become an official discord lib

- [x] **Stage 1.** Accept `EC_SHARD_ID` and `EC_SHARD_NUM` as environment variables and send these to Discord. The user deploying the bot is responsible for configuring each instance of the bot correctly. #120 
- [ ] **Stage 2.** Connect to the [Get Gateway Bot](https://discordapp.com/developers/docs/topics/gateway#get-gateway-bot) endpoint to retrieve the recommended number of shards. Based upon that result setup N shards per host. For example, with `EC_SHARD_NUM=2` and a recommendation of 5 shards from the endpoint, host 1 would connect as shards 0,1,2, with host 2 connecting as shards 3,4.
- [ ] **Stage 3.** Using a shared/clustered data store determine how many instances of the bot are running to work out `EC_SHARD_NUM` and `EC_SHARD_ID` automatically. Then setup N shards per host as per step 2.
- [ ] **Stage 4.** Automated shard addition and removal. Using the data store as in Stage 3 monitor factors that could affect the number of shards we need to have. e.g., the recommendation from discord, the number of hosts running, hosts going down etc.

PR #166, opened 2018-06-12
Addresses multiple intertwining issues.

CoreyShupe:
Fixes https://github.com/princesslana/ErisCasper.java/issues/71
Fixes https://github.com/princesslana/ErisCasper.java/issues/120
Addresses https://github.com/princesslana/ErisCasper.java/issues/105
Addresses https://github.com/princesslana/ErisCasper.java/issues/122

The reason for these all being included in this PR is they all have to do with connection and really they go hand-in-hand.


src/main/java/com/github/princesslana/eriscasper/BotContext.java
```diff
@@ -19,13 +22,19 @@
...
+  private final ErisCasper erisCasperInstance;
   private final Observable<Event> events;
 
   private Routes routes;
 
   private RepositoryManager repositories;
 
-  public BotContext(Observable<Event> events, Routes routes, RepositoryManager repositories) {
+  public BotContext(
+      ErisCasper erisCasperInstance,
```
ianagbip1oti: We should need to keep an ErisCasper instance in BotContext

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -26,41 +30,107 @@
   private final ObjectMapper jackson = Jackson.newObjectMapper();
   private final Payloads payloads = new Payloads(jackson);
 
+  private final Integer[] shard;
```
ianagbip1oti: Let's use a class provides more semantics than `Integer[]`

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -26,41 +30,107 @@
...
         .flatMap(routes::execute)
         .toObservable()
-        .flatMap(gr -> gateway.connect(gr.getUrl(), token))
+        .flatMap(gr -> gateway.connect(gr.getUrl(), token, shard))
         .observeOn(Schedulers.computation())
         .share();
   }
 
+  public Integer getShardNum() {
+    return shard == null ? -1 : shard[0];
```
ianagbip1oti: Why return -1 when we could use `Optional`?

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -26,41 +30,107 @@
...
+  }
+
+  public Completable updateVoiceState(UpdateVoiceState update) {
+    return gateway.updateVoiceState(update);
+  }
+
   public static ErisCasper create() {
-    return create(System.getenv("EC_TOKEN"));
+    String token;
+    if (System.getenv().containsKey("EC_TOKEN")) {
```
ianagbip1oti: Using custom and varied handling of system properties/environment variables is just going to be more code we have to change when #106 gets implemented.

I'd very much prefer to use environment variables only until #106 is done.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -26,41 +30,107 @@
...
+      token = System.getenv("EC_TOKEN");
+    } else {
+      token = System.getProperty("ec.token", null);
+    }
+    if (token == null) {
+      throw new ErisCasperFatalException("Token cannot be null upon creation.");
+    }
+    String shardNumRaw = System.getProperty("shard.num", null);
+    String shardTotalRaw = System.getProperty("shard.total", null);
+    if (shardNumRaw != null
```
ianagbip1oti: This kind of validation logic is what can be shifted into a Shard class

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -26,41 +30,107 @@
...
+    bot.apply(new BotContext(this, events, routes, rm))
         .doOnError(t -> LOG.warn("Exception thrown by Bot", t))
         .blockingAwait();
   }
 
+  public void shutdownGracefully() {
+    gateway.shutdownGracefully();
+  }
+
+  public Completable requestGuildMembers(RequestGuildMembers request) {
```
ianagbip1oti: This looks like a candidate for a repository somewhere.

When the client wants to fetch data, they should go via repositories. When they want to do something, it should be an Action.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/RequestGuildMembers.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @see <a href="https://discordapp.com/developers/docs/topics/gateway#request-guild-members">
+ *     https://discordapp.com/developers/docs/topics/gateway#request-guild-members</a>
+ */
+@Value.Immutable
+public interface RequestGuildMembers {
+  @JsonProperty("guild_id")
+  Snowflake getGuildId();
+
+  @JsonProperty("query")
```
ianagbip1oti: Be consistent with how we specify these `@JsonPropery` annotations. Either we always specify them, or we only specify them when we need to.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/UpdatePresence.java
```diff
@@ -0,0 +1,27 @@
...
+import org.immutables.value.Value;
+
+/**
+ * @see <a href="https://discordapp.com/developers/docs/topics/gateway#update-status">
+ *     https://discordapp.com/developers/docs/topics/gateway#update-status</a>
+ */
+@Value.Immutable
+public interface UpdatePresence {
+  @JsonProperty("since")
+  @Nullable
```
ianagbip1oti: Why `@Nullable` instead of `Optional`?

src/main/java/com/github/princesslana/eriscasper/gateway/commands/util/ConnectionProperties.java
```diff
@@ -0,0 +1,29 @@
+package com.github.princesslana.eriscasper.gateway.commands.util;
```
ianagbip1oti: Why is this in its own package?

`util` is often a bit of a meaningless name because it can mean anything. Often it's a dumping spot because there wasn't a better name. If this is the case, wouldn't it be better off near the class that uses it?

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -29,31 +31,63 @@ public RxWebSocket(OkHttpClient http) {
...
 
-              ws = http.newWebSocket(rq, new Listener(em));
+              ws = http.newWebSocket(rq, new Listener(RxWebSocket.this, em));
             })
+        .takeUntil(event -> closed)
         .doOnNext(e -> LOG.trace("Received: {}.", e))
         .doOnError(e -> LOG.warn("Error: {}.", e));
   }
 
+  private void close0() {
```
ianagbip1oti: Why `close0`? Does this method name make it obvious what this method does?

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -29,31 +31,63 @@ public RxWebSocket(OkHttpClient http) {
...
     }
 
     @Override
     public void onClosing(WebSocket ws, int code, String reason) {
+      if (!socket.closed) {
+        if (code == 4004) {
+          socket.close(1002, "Invalid token.");
+          new ErisCasperFatalException("Failed to authenticate with discord servers.")
+              .fillInStackTrace()
+              .printStackTrace();
```
ianagbip1oti: Don't print to stdout.

Why `fillInStackTrace`? Did you mean to `throw` this exception?

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -29,31 +31,63 @@ public RxWebSocket(OkHttpClient http) {
             em -> {
               Request rq = new Request.Builder().url(url).build();
 
-              ws = http.newWebSocket(rq, new Listener(em));
+              ws = http.newWebSocket(rq, new Listener(RxWebSocket.this, em));
             })
+        .takeUntil(event -> closed)
```
ianagbip1oti: Referring to mutable things outside of the reactive stream is a bit of a code smell - it's kind of like the global variables of rx.

ianagbip1oti (changes requested):

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: This isn't fetching members. This is asking for offline guild members to be noticed in very large guilds. The naming is a bit weird for this.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/UpdatePresence.java
CoreyShupe: Passing through null afaik means something different than not passing it through at all.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/util/ConnectionProperties.java
CoreyShupe: We can put it as an inner class to Identify.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: it's the default closing of the socket private to the class. it's the ultimate goto when wanting to close the connection and stop accepting packets.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: throwing it leaves a long very ambiguous looking error, seems to leave a much cleaner looking error if we directly print it, more useful to the end user imo, but I can change it.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: I think I can remove that, I think those are in there for no reason at this point since I added in the basic level of `takeUntil`

src/main/java/com/github/princesslana/eriscasper/gateway/commands/RequestGuildMembers.java
CoreyShupe: not sure of when we need them, when are the instances where we do need them? I'm not too sure on the circumstance.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/UpdatePresence.java
CoreyShupe: I'll look into it and see if it's optional, but from what I saw it's necessary but nullable

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: Hmm looked at it a bit more, the reactive stream needs to look outside of the reactive stream to check if the RxWebsocket is closed before it can continue. Is there a good reason we should change this? It makes the most sense for shutting down the stream so we stop sending heartbeats and such when the pipe breaks or the socket closes.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: removed this alltogether, we only need the shard for identifying

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
ianagbip1oti: ok, it would still seem to be better in a repository as `getOfflineGuildMembers`.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/RequestGuildMembers.java
ianagbip1oti: Jackson by default converts by remove the `get` and lower casing the initial letter.

So, `getQuery` looks at the `query` field. `getGuildId` would look at the `guildId` field by default, which doesn't match with `guild_id` that we want.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
ianagbip1oti: If we can't continue, then throwing an exception is the best way to indicate this

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: Maybe but I'd much rather support it the same as all the other gateway commands, it doesn't return anything, it just sends `GuildMemberChunkEvent`s so it's not like it returns anything, it just sends a request for events to update guild members. We could throw some sort of implementation in the repo.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
ianagbip1oti: We should detect closing by seeing the Closed websocket event: https://github.com/princesslana/ErisCasper.java/blob/master/src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocketEvent.java#L14

src/main/java/com/github/princesslana/eriscasper/gateway/commands/RequestGuildMembers.java
CoreyShupe: ah, alright, I'll go through and modify the json properties I set to match with that

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: Alright, fair enough, I'll throw it rather than printing it.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: gotcha, implemented that

src/main/java/com/github/princesslana/eriscasper/api/robot/Robot.java
```diff
@@ -68,8 +68,8 @@ public void respond(String regex, Function<RobotContext, Completable> f) {
    * Checks all messages sent directly to the Robot against the provided regex. Executes the
    * provided function when there is a match.
    *
-   * <p>Messages are considered to be directed at the Robot if they begin with "+", the username of
-   * the bot account, or a mention of the bot account.
+   * <p>Messages are considered to be directed at the Robot if they begin with "+", the username on
+   * the bot account, or a mention on the bot account.
```
ianagbip1oti: why are all these 'of's replaced with 'on'. I don't think it's correct in all cases.

CoreyShupe: Yeah I don't actually know when I did that.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/UpdatePresence.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @see <a href="https://discordapp.com/developers/docs/topics/gateway#update-status">
+ *     https://discordapp.com/developers/docs/topics/gateway#update-status</a>
+ */
+@Value.Immutable
+public interface UpdatePresence extends GatewayCommand {
+  @JsonProperty("since")
+  @Nullable
+  Integer getSince();
```
ianagbip1oti: Reading the docs to me seems like this can be absent, or present and null.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/UpdateVoiceState.java
```diff
@@ -0,0 +1,56 @@
...
+
+  static UpdateVoiceState disconnect(Snowflake guildId) {
+    return connect(guildId, null);
+  }
+
+  static UpdateVoiceState connect(Snowflake guildId, Snowflake channelId) {
+    return connect(guildId, channelId, false, false);
+  }
+
+  static UpdateVoiceState connect(
```
ianagbip1oti: I think this starts to get too much. I think the client should just use the builder themselves instead of a four parameter method. Method signatures like (Snowflake, Snowflake, boolean, boolean) are error prone also. It's easy to accidentally flip the order of parameters.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/UpdatePresence.java
CoreyShupe: Having the option for the user to have it as null is important, Optional restricts that by leaving it out. I don't think leaving it out and having it null have the same effect.

src/main/java/com/github/princesslana/eriscasper/gateway/commands/UpdateVoiceState.java
CoreyShupe: Gotcha, I'll remove the connect/disconnect methods, I thought it might be too much as well.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -16,6 +17,7 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(RxWebSocket.class);
 
+  private volatile boolean closed = false;
```
ianagbip1oti: This is to be removed in favor of picking up the Closed web socket event?

CoreyShupe: I need to see where else it's being used, if unused I'll remove it.

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
```diff
@@ -0,0 +1,49 @@
+package com.github.princesslana.eriscasper.util;
+
+import com.github.princesslana.eriscasper.ErisCasperFatalException;
+import com.github.princesslana.eriscasper.data.immutable.Wrapper;
+import com.ufoscout.properlty.Properlty;
+import java.util.Optional;
+
+public class Shard implements Wrapper<Integer[]> {
```
ianagbip1oti: I don't think implementing Wrapper here is semantically correct. I think just having a `toArray` method is cleaner.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: Yeah it's being removed

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
```diff
@@ -0,0 +1,49 @@
...
+    this.shardNumber = shardNumber;
+    this.shardTotal = shardTotal;
+  }
+
+  @Override
+  public Integer[] unwrap() {
+    return new Integer[] {getShardNumber(), getShardTotal()};
+  }
+
+  public int getShardNumber() {
```
ianagbip1oti: Do we need an unwrap method along with the getters? Seems like one or the other should be redundant.

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
ianagbip1oti: Have you considered using immutables and making this a `@Tuple`? You could still do the checks on the values by creating our own `Shard.of` factory method.

CoreyShupe: Implementing Wrapper allows it to be used with jackson properly. It also is correct in the sense that it's wrapping an array of integers. If I don't use Wrapper Jackson may not like it so much. If you know how to get it to where Jackson treats it like an Integer[] like it does with Wrapper I can replace it.

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
CoreyShupe: Yeah I'll probably remove the getters.

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
```diff
@@ -0,0 +1,49 @@
...
+import com.ufoscout.properlty.Properlty;
+import java.util.Optional;
+
+public class Shard implements Wrapper<Integer[]> {
+
+  private final int shardNumber;
+  private final int shardTotal;
+
+  public Shard(int shardNumber, int shardTotal) {
+    if (shardNumber >= shardTotal || shardTotal < 1 || shardNumber < 0) {
```
ianagbip1oti: `Preconditions.checkArgument` from Guava works well for this kinds of checks. It'll throw `IllegalArgumentException` instead of `ErisCasperFatalException`. I can see arguments for either.

CoreyShupe: I don't think it's necessary to mix exceptions. It's also a very primitive thing to check, I don't think it's necessary to wrap it in Preconditions.

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
ianagbip1oti: Except we don't wrap an array of integers, we have two integer fields.

The Jackson magic you are looking for is `@JsonValue`. If you annotate the `toArray` method with it, then Jackson will use that to serialize.

CoreyShupe: Alright, I'll use that then.

src/main/java/com/github/princesslana/eriscasper/BotContext.java
```diff
@@ -37,13 +38,17 @@ public BotContext(Observable<Event> events, Routes routes, RepositoryManager rep
...
-    return events.ofType(evt).map(e -> e.unwrap()).flatMapCompletable(f::apply);
+    return events.ofType(evt).map(E::unwrap).flatMapCompletable(f::apply);
   }
 
   public Completable doNothing() {
     return Completable.complete();
   }
 
+  public Completable sendGatewayCommand(GatewayCommand command) {
+    return routes.sendGatewayCommand(command);
```
ianagbip1oti: I'm not sure this belongs on Routes. Routes is kind of HTTP specific.

But for what reason are we exposing the gateway to the user?

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -38,21 +40,22 @@
   private final Payloads payloads = new Payloads(jackson);
 
   private final Routes routes;
+  private final Optional<Shard> shard;
 
-  private ErisCasper(BotToken token) {
+  private ErisCasper(BotToken token, Optional<Shard> shard) {
```
ianagbip1oti, line 45: Should we avoid this optional by having using a Shard of [0,1] as a default?

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -85,14 +88,31 @@ private void sequenceNumberSeen(Optional<SequenceNumber> seq) {
...
     Observable<Payload> ps =
         ws.connect(String.format("%s?v=%s&encoding=%s", url, VERSION, ENCODING))
             .ofType(RxWebSocketEvent.StringMessage.class)
             .map(RxWebSocketEvent.StringMessage::getText)
             .flatMapMaybe(
                 Singles.toMaybeAnd(
                     payloads::read, (s, t) -> LOG.warn("Error reading payload: {}", s, t)))
+            .takeUntil(
+                payload -> {
+                  if (payload.op().getCode() == 9) {
```
ianagbip1oti: Do we not have an enum or constants around instead of having to use 9 here?

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -146,14 +166,22 @@ private Completable resume(RxWebSocket ws, BotToken token) {
...
 
-    return Single.just(
-            ImmutableResume.builder()
-                .token(token)
-                .sessionId(sessionId.get())
-                .seq(lastSeenSequenceNumber.get())
-                .build())
-        .map(payloads::resume)
-        .flatMapCompletable(p -> send(ws, p));
+    return send(
```
ianagbip1oti: I'm 50/50 about this send method.

It does reduce a bit of duplication, but it also makes the method hard to read.

With the flatMapCompletable it reads in sequence: build the resume payload, apply/map the payloads resume function, and then send it. The send at the start kind of has it out of sequence a bit.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -29,21 +30,39 @@ public RxWebSocket(OkHttpClient http) {
...
 
+  public void close(int code) {
+    close(code, null);
+  }
+
+  public void close(int code, String reason) {
+    ws.close(code, reason);
+  }
+
+  public Completable closeDueToInvalidSession() {
```
ianagbip1oti: This is a leaky abstraction. Invalid session is a discord concept. The RxWebSocket class shouldn't care whether it's being used in discord or not.

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
ianagbip1oti: The judgement should be made based on readability. I agree it's not a big deal with exception is throw and that the code is straight forward. Either changing or leaving is fine.

```
Preconditions.checkArgument(shardNumber >= 0, "shardNumber should be zero or more");
Preconditions.checkArgument(shardTotal >= 1, "shardTotal should be one or more");
Preconditions.checkState(shardNumber < shardTotal, "shardNumber should be less than shardTotal");
```

src/main/java/com/github/princesslana/eriscasper/BotContext.java
CoreyShupe: Gateway commands must be sent under the gateway, I dont know of a way to format them under a request since they have no actual statement. If we want to let users send commands to the gateway we need to expose it somehow. Safest way to do that is to give them access to a method which sends the command for them. I can remove Resume and Identify from GatewayCommand and it would make it a little safer in that regard.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: I think having no sharding is a bit safer, I dont think telling discord we are using shard [0,1] is useful and Optional makes sense and looks like what it does.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: I'll change it

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: I can rename it, i think that's the best thing to do in this situation.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: yeah you're right, I'll remove it and push it into gateway

src/main/java/com/github/princesslana/eriscasper/util/Shard.java
CoreyShupe: I might change it to preconditions then I do think that is more descriptive, tests will change with it

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: I'm thinking `completePayload`

src/main/java/com/github/princesslana/eriscasper/BotContext.java
ianagbip1oti: When the user wants to do something like send a message, the fact that it's happening over HTTP is hidden from them. There is nothing in the API that refers to rest or http, they do it using the `Action` concept we have.

So, gateway commands can be done in the same way.

Perhaps `Action` is an interface, with implementations `RestAction`, `GatewayAction`.

CoreyShupe: That's kind of impractical if we want the users to know which one to use, we also need to know which one it is because each of them have special attributes. It would cause casting and checking and such if we just wanted to use Action.

CoreyShupe:
@ianagbip1oti are we holding back this PR until Data is pretty complete with the update we're doing. Or is there more comments you have for it?

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
ianagbip1oti: It's not the naming that I'm concerned about it's the flow. By having the send in the map at the end, you end up with the code reading as: start with something, do this, do that, do this, then send it. With the method you end up with: send (something, do this, do that, do this )

ianagbip1oti:
There are still open comments on this. e.g., the exposing of 'gateway' to the user

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -85,14 +88,31 @@ private void sequenceNumberSeen(Optional<SequenceNumber> seq) {
...
+                  if (payload.op() == OpCode.INVALID_SESSION) {
+                    if (payload.d().map(JsonNode::asBoolean).orElse(false)) {
+                      resume(ws, token);
+                    } else {
+                      // blocking since we shouldn't be doing anything if this happens
+                      // (this is an invalid session which can't be resumed.
+                      Completable.fromAction(() -> ws.close(1002, "Invalid session."))
+                          .doOnComplete(
+                              () -> LOG.warn("Socket disconnected due to an invalid session."))
+                          .blockingAwait();
```
ianagbip1oti: We shouldn't need to use something blocking during an observable stream.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -54,6 +66,12 @@ public void onClosed(WebSocket ws, int code, String reason) {
 
     @Override
     public void onClosing(WebSocket ws, int code, String reason) {
+      if (code == 4004) {
+        socket.close(1002, "Invalid token.");
+        em.onNext(ClosingTuple.of(ws, code, reason));
+        throw new ErisCasperFatalException(
+            "Failed to authenticate with discord servers: [" + reason + "]");
```
ianagbip1oti: The websocket class shouldn't know about the existence of discord.

Why can't the discord listeners listen for the Closing event and handle this?

CoreyShupe: Didn't even realize this was still in the codebase. I'll move it.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: It's blocking because we don't want anything more to even try to come through, if we do it will throw broken pipes and have issues with the websocket. If we block it then we can avoid anything even attempting to come through when it's a non-resumable invalid session.

CoreyShupe:
The gateway isn't exposed anymore? It's just mixing `execute` with taking more types of parameters. Once the Data gets pushed through it may be more viable to set up `route` type things with gateway commands and just have routes be able to handle gateway routes vs http routes. Or maybe it would be smarter to create an intervening class which handles the different and handles it the way it needs to be handled, for example if it's an http route do this, if it's a gateway route do this. So ChannelRoute and GuildRoute etc.. would implement something like `HttpRoute` and we can have `GouldRoute` and an interface `Routed` with a `getType` method to handle the special conditions of it. That might be the smartest thing to do in this situation. Although I am open to suggestions.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
ianagbip1oti: Putting the blocking here isn't stopping other websocket events from coming through or happening.

CoreyShupe: Alright, how would we push through that command into the stream? Should we just run it outside of a completable.

ianagbip1oti: I'm not really sure, this isn't something I have thought about. I'm not 100% clear on what situations lead to this happening and what action we're taking.

ianagbip1oti:
The eventual goal is that all requests for data go via repositories. So, if I want to get all guild members I get a repository via the BotContext and call`getGuildMembers`. Then, I shouldn't care whether it's going via the gateway, reading from something cached in memory, or via the REST API.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: I'm not exactly sure what _can_ throw a non-resumable invalid sessions in all honesty. I tried for around 3 hours to get github to throw me an invalid session without success. Invalid tokens throw an initial unauthorized message not an invalid session message so that's a bit odd. Basically right now within this codebase we're just shutting down the websocket.

ianagbip1oti: Ok, so I'm not keen to merge something where we're not sure what we're meant to be doing.

CoreyShupe:
So really the goal is pushing all route commands / gateway commands into repositories? However the issue I see with that is most gateway commands don't return data. They aren't anything for repositories to care about, just because they are requests/commands made to the discord API to perform specific actions. Same with certain route commands. Like, prune messages, or update voice state. I'm not sure how those are at all related to repositories so I find it more into the thought of building a command API we can throw into bot context to reach all of those commands without the user knowing about the gateway or our rest API. Eventually we should never see Route<X, Y> as the user and instead be entirely concerned about Actions. Actions need to be reformed though since they don't at all accept special conditions.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: Alright, maybe we should remove it then, or just throw a fatal exception when I receive an invalid session which can't be resumed. Most invalid sessions are able to be resumed and I haven't seen one where I couldn't. So maybe that's the right move? Not 100% sure.

ianagbip1oti:
Ok, to clarify, I didn't mean all gateway commands (and http routes) should be in repositories.

Repository - where you query something about discord or the state
Action - what you use to do something

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
ianagbip1oti: I'd say remove anything related to #64 from this PR and address it at another time.

CoreyShupe:
Alright, so the `getGuildMembers` thing you're referring to is basically telling discord `I want to know more about this guild, I want to know the offline members as well, so send me GuildMemberChunk events until I have known about every single offline member` to which the repository already takes care of. So it's less of a query for information which receives and more of a telling discord to send me more events. It could be seen as a repository request but the issue is it isn't returning anything or telling us anything, it's just making discord reveal more information which will be available later.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: Sounds good to me. I think it's a more complicated issue due to being able to test and understand the circumstances that lead to the outcomes we're trying to prevent.

ianagbip1oti:
getGuildMembers should be in a repository.

The say I see it is it should handle the GuildMemberChunk events as they come in, and present them as something like `Observable<GuildMember>` back to the user.

CoreyShupe:
My only hesitation comes when the user doesn't understand the command. `getGuildMembers` returns offline members when and only when the guild has over 100 members inside of it. If it has less than 100 it will return nothing. Also, what if I call that method for the same guild multiple times? It will create 2 observables and when the events come in how will I know which observable to throw it into? The only thing I'm trying to prevent with this is user based error due to ignorance of the backend mechanics. `which will prevent issues being reported when none are based off issues`

ianagbip1oti:
ErisCasper should take care of how getGuildMembers happens. So, ErisCasper should know whether it should get them via a gateway event, or via a REST call. Or maybe it just returns what it has cached in memory. This is why we want to put it behind the repository interface - so that the bot writer doesn't have to care about that.

CoreyShupe:
Then maybe we should implement the gateway command into the initialization of the repository so the user doesn't ever have to care about it and we automagically cache every single guild member anyways. When a guild is presented we run the get guild members gateway command and it throws in chunk events so the user is always able to access every guild member.

ianagbip1oti:
I'd prefer it to be lazy/on-demand

CoreyShupe:
I see it more complicated than that. Just because from the gateway perspective.
The command comes in -> send events back to the user -> cache the guild members.
Should we present getGuildMembers as instead of a command just a return of the cached guild members? That might cause issues with people who want all guild members including offline ones, so maybe we should have something on the initialization defining if they want to see offline members? I'm not really sure how we should do this because the gateway command is moreso for telling discord we want more not that we have more to give to the user.

ianagbip1oti:
This is why I like things as separate PRs. Because it allows more thinking and reviewing of things without blocking other changes.

From the bot writers perspective, I should call `getGuildMembers` on a repository. This should give me the result with the principle of least surprise (i.e., it should give me all members of the guild).

How we implement that, is a separate discussion, and yeah, there are issues to work through like you raised (e.g., if we cache, how do we know the cache is valid?).

But, Observable is a great abstraction for presenting the GuildMembers back to the bot writer. They write code that operators on the Observable, that is called when the GuildMembers become available - it doesn't expect them to all be there immediately.

CoreyShupe:
Alright fair enough. In all honesty, I may strip gateway commands entirely from here considering we're handling them through Data and representing them here in multiple different ways. I don't think gateway commands can be handled in the current state of the codebase. I think we should first reformat the Actions API and the repositories to handle information better and have a more viable state. I think right now we're moreso in the stage of setting up the skeletal framework and the framework for gateway commands is being done in -Data now.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -38,21 +40,23 @@
   private final Payloads payloads = new Payloads(jackson);
 
   private final Routes routes;
+  private final Optional<Shard> shard;
 
-  private ErisCasper(BotToken token) {
-    this.token = token;
+  private Gateway gateway;
```
ianagbip1oti: final? </nit>

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -29,21 +29,32 @@ public RxWebSocket(OkHttpClient http) {
...
+  }
+
   public Completable send(String text) {
     return Completable.fromAction(() -> ws.send(text))
         .doOnComplete(() -> LOG.trace("Sent: {}.", text));
   }
 
   private static class Listener extends WebSocketListener {
     private final ObservableEmitter<RxWebSocketEvent> em;
+    private final RxWebSocket socket;
```
ianagbip1oti: This is unused?

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -29,21 +29,32 @@ public RxWebSocket(OkHttpClient http) {
...
 
-              ws = http.newWebSocket(rq, new Listener(em));
+              ws = http.newWebSocket(rq, new Listener(RxWebSocket.this, em));
             })
+        .takeUntil(e -> e instanceof RxWebSocketEvent.Closed)
         .doOnNext(e -> LOG.trace("Received: {}.", e))
         .doOnError(e -> LOG.warn("Error: {}.", e));
   }
 
+  public void close(int code) {
```
ianagbip1oti: Is this used?

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: the gateway is initialized later, not when created

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: yeah I think I can remove that now

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: I can remove this, yeah, leftover

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
ianagbip1oti: Is gateway used out side the `getEvents` method?

CoreyShupe: I'll probably remove the field altogether and add it with the action rewrite

CoreyShupe:
@ianagbip1oti this needs to be commented on.
The scope of this now addresses how we're doing gateway commands.
Implements a sharding technique (which I think will be replaced by 0.6.0 data classes) but is useful for a placeholder until we move the API object into Data. (fixed stage 1 and addresses the entire issue)
Fixes bad tokens.

ianagbip1oti:
Which is the best path forward to 0.6.0 data? Making a PR to upgrade to data 0.6.0 and merging that before updating this, or merging this and then modifying what's here for data 0.6.0?

CoreyShupe:
I think the most logical is merging this then modifying. If we update to 0.6.0 then merge the PRs might get messy. Once 0.6.0 is updated we need to make everything conform anyways. There's going to be quite a bit in the update PR just because of how much changed between 0.5.0 -> 0.6.0. Not much is changing within this PR besides the sharding pieces afaik. This also kind of keeps things consistent with time and commit consistency for everyone. Neater.

ianagbip1oti (approved):

Issue #108, opened 2018-03-25
Documentation


WangzoR:
Overview of content and supporting pages that need to be created for and linked

When starting work on the project, please go through the following.

Must Have

- [ ]     Project Overview/Oversight Page?
                Key milestones?
                Targets?
                What is classified as MVP for official discord status consideration?
                https://discordapi.com/unofficial/comparison.html?
                Outstanding epics?

Must Have

- [ ] 	   Install the project on your computer (links to "Setting up the project locally page")
		Access
		Software/Framework/Lib Dependencies
		Source file locations/info
		Setup Guide per IDE/environment 
		Setup Common Pitfalls
		Localised running/testing?

- [ ]     Read up on the project (links to "About the Project page(s)")
		Discord
		DiscordAPI
		Vision/Goal
		ReactiveX 
		Architecture Overview
		Milestones/Phases
		Timeline?
		Project Goverance/Structure
		Team/Go To People/Contributors
		
Should Have   

- [ ] 	If you are not familiar with ReactiveX/Java or could use a refresher
                Please make sure to familiarise yourself ( Links to externally hosted ReactiveX guides)
                Link to official ReactiveX pages
		Tutorials/Guides
		Examples
		
- [ ] 	Read and understand the coding guidelines / standards  (Links to "Coding Standards")
		White Space
		Naming Convention
                Java Style Guide

Could Have    

- [ ] 	Read and understand the development process (Links to "Development & Work Process")
		Logging Issues Guidelines
		Picking Up Issues and progression Guidelines
		Labels Usage
		Branching Methodology
		Pull Request Guidelines
		Unit testing
	
Reference material
![reimport-maven-intellij](https://user-images.githubusercontent.com/37465280/37871666-172a9246-303f-11e8-9423-4d9fd10faf9f.png)
https://github.com/twitter/commons/blob/master/src/java/com/twitter/common/styleguide.md
http://reactivex.io/
http://reactivex.io/intro.html	
https://princesslana.github.io/ErisCasper.java/
https://gist.github.com/meew0/bbbbd5348967dee5f7e84c0cd58983fd
https://immutables.github.io/apt.html
https://github.com/princesslana/ErisCasper.java/tree/master/src/site/markdown
https://github.com/princesslana/ErisCasper.java/blob/master/src/site/site.xml#L9
https://github.com/WangzoR/ErisCasper.java/blob/master/ISSUE_TEMPLATE.md
[...]

PR #167, opened 2018-06-14
Support system property configuration

### LINKED ISSUE #106: Accept systems properties as well as environment variables
Currently ErisCasper accepts configuration via environment variables (e.g., `EC_TOKEN=my.bot.token`).

Java projects often respect system properties as well (e.g., `-Dec.token=my.bot.token`)

We should accept either. If both are given, prioritize the environment variable setting.

ianagbip1oti:
Fixes #106

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
+++ b/src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
@@ -9,6 +9,10 @@ import com.github.princesslana.eriscasper.repository.RepositoryManager;
 import com.github.princesslana.eriscasper.rest.RouteCatalog;
 import com.github.princesslana.eriscasper.rest.Routes;
 import com.github.princesslana.eriscasper.util.OkHttp;
+import com.ufoscout.properlty.Properlty;
+import com.ufoscout.properlty.reader.EnvironmentVariablesReader;
+import com.ufoscout.properlty.reader.SystemPropertiesReader;
+import com.ufoscout.properlty.reader.decorator.ToLowerCaseAndDotKeyReader;
 import io.reactivex.Observable;
 import io.reactivex.Single;
 import io.reactivex.schedulers.Schedulers;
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -57,7 +68,10 @@ public void run(Bot bot) {
   }
 
   public static ErisCasper create() {
-    return create(System.getenv("EC_TOKEN"));
+    return create(
+        CONFIG
+            .get("ec.token")
```
CoreyShupe, line 73: How does this work? Should we throw in a default configuration file for example. I think this is a bit ambiguous as to what to do. Maybe throw a config file into the resources and if not built create a default one for them. Something needs to be done so before installation there's a way to figure out how to set it up, rather that be in docs or some other example.

CoreyShupe (changes requested):

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
ianagbip1oti: This doesn't read from files yet, it just adds system properties as per #106. So you can configure the token by setting the `EC_TOKEN` environment variable, or the `ec.token` system property.

We do need a getting started guide, but imo that's outside the scope of this issue.

CoreyShupe: Ah okay, so this is just the beginning to implementing a configuration file, or rather the implementation to provide support for it.

ianagbip1oti: It's the beginning of enabling configuration to happen in ways other than just environment variables, yes.

CoreyShupe: Then this is probably good for now, it definitely needs work in documentation and examples for installation. I'll put that in the issue list.

ianagbip1oti: #108 exists for documentation

CoreyShupe: I don't see configuration specification. Should we add that as a checklist item.

ianagbip1oti: Or leave a comment

CoreyShupe: Alright, I'll leave a comment there and accept this PR through.

CoreyShupe (approved):

PR #169, opened 2018-06-19
Remove unused parameters in Route methods

ianagbip1oti:
Fixes:
* https://sonarcloud.io/project/issues?id=com.github.princesslana%3AErisCasper.java&issues=AWOF4j4Prr7mtNquU9AK&open=AWOF4j4Prr7mtNquU9AK
* https://sonarcloud.io/project/issues?id=com.github.princesslana%3AErisCasper.java&issues=AWN13uUPTmDVX2bf0Rb8&open=AWN13uUPTmDVX2bf0Rb8
* https://sonarcloud.io/project/issues?id=com.github.princesslana%3AErisCasper.java&issues=AWN13uUPTmDVX2bf0Rb9&open=AWN13uUPTmDVX2bf0Rb9

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/rest/ChannelRoute.java
+++ b/src/main/java/com/github/princesslana/eriscasper/rest/ChannelRoute.java
@@ -33,7 +33,7 @@ public class ChannelRoute {
    *     https://discordapp.com/developers/docs/resources/channel#modify-channel</a>
    */
   public Route<ModifyChannelRequest, Channel> modifyChannel() {
-    return Route.put(path("/"), ModifyChannelRequest.class, Channel.class);
+    return Route.put(path("/"), Channel.class);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


CoreyShupe (commented):
Could this be done with every single route type so we don't even need to define classes, I'm pretty sure generic `guessing` will work well with this. I doubt we need to define the classes when looking for a specific generic set. Looks like what we've done is half of what we could do.

ianagbip1oti:
What do you mean by generic guessing?

CoreyShupe:
In the past when I've used generics you don't need to define types in methods if you're returning a generic set which I'm requesting. For example.
```java
public <X, Y> Obj<X, Y> get() {
  return Obj.of();
}

// will work for

public Obj<Integer, Boolean> getIntBool() {
  return get();
}
```
that afaik only works because you know what you're requesting down the line. Looks nicer and we don't need to define return types, unless we're trying to really define the function. Just makes it so we don't have to use the `X.class` and just use raw generics, pretty sure the JDK and most IDEs should understand that I want an Integer and a Boolean from int and auto assign.

ianagbip1oti:
The reason that doesn't work is that we do need X.class - we need to pass it to Jackson so that it knows how to create an instance of the class it will deserialize to.

CoreyShupe:
Yup, you're right. Didn't notice that at first. Hence why it's more of a comment and not a requested change. Was unsure of if it was needed or not.

CoreyShupe (approved):

PR #171, opened 2018-06-30
Add core API page

ianagbip1oti:
An attempt at documenting core api concepts.

### DIFF
--- a/docs/core_api.md
+++ b/docs/core_api.md
@@ -2,7 +2,7 @@
 
 The core API of ErisCasper.java is designed to be minimal and powerful.
 
-It exposes the concepts of `Bot`s, `Action`s, and `Repository`s as
+It exposes the concepts of `Bot`s, the event stream (`Observable<Event>`), `Action`s, and `Repository`s as
 described below.
 
 There also exists the concept `Gateway` and `Routes`.
@@ -16,13 +16,21 @@ developers if required.
 
 A `Bot` is a `Function` run with access to a `BotContext` and produces a `Completable`.
 
-The incoming event stream, `Repository`s, and ability to create `Action`s are 
+The event stream, `Repository`s, and ability to create `Action`s are 
 accessible via the `BotContext`.
 
 The return type is a `Completable` as defined by RxJava.
 This Completable will be subscribed to by ErisCasper.java.
 
 `Bot`s can be composed. Multiple implementations of `Bot` may be merged into a single `Bot`.
+
+
+## Event Stream
+
+The event stream is the events sent from Discord.
+This includes events such as message sends, new guild members, etc.
+
+The event stream uses the `Observable` type from RxJava to make the events available to bot developers.
 
 
 ## Actions
--- a/docs/core_api.md
+++ b/docs/core_api.md
@@ -1,0 +1,53 @@
+# Core API
+
+The core API of ErisCasper.java is designed to be minimal and powerful.
+
+It exposes the concepts of `Bot`s, `Action`s, and `Repository`s as
+described below.
+
+There also exists the concept `Gateway` and `Routes`.
+These are lower level wrappers around the Discord API and considered internal
+to ErisCasper.java.
+This means their use is heavily discouraged, but they are available to bot
+developers if required.
+
+
+## Bot
+
+A `Bot` is a `Function` run with access to a `BotContext` and produces a `Completable`.
+
+The incoming event stream, `Repository`s, and ability to create `Action`s are 
+accessible via the `BotContext`.
+
+The return type is a `Completable` as defined by RxJava.
+This Completable will be subscribed to by ErisCasper.java.
+
+`Bot`s can be composed. Multiple implementations of `Bot` may be merged into a single `Bot`.
+
+
+## Actions
+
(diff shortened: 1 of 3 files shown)


CoreyShupe:
Should we address the events and how they're done?

ianagbip1oti:
I've added the "Event Stream" as a concept in the core api.

After thinking about it Events, Resources, Requests, etc are really just data that is passed around through the api. These could be mentioned in this doc, but then I think it starts to lose a little bit of focus. The focus currently is on the APIs that allow you to do stuff, without getting into specifics of what data you need to pass around and receive.

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
...
+but for which there is no response expected.
+Examples include sending a message, or adding a role.
+
+
+## Repositories
+
+`Repository`s are the source of information from Discord.
+This could include, for example, information about guilds, channels, or users.
+
+`Repository` implmentations may perform caching, gateway requests, or HTTP requests
```
CoreyShupe: Spelling error.
See http://www.dictionary.com/browse/implementation
"implmentations"

CoreyShupe (changes requested):

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
+# Core API
+
+The core API of ErisCasper.java is designed to be minimal and powerful.
+
+It exposes the concepts of `Bot`s, the event stream (`Observable<Event>`), `Action`s, and `Repository`s as
+described below.
+
+There also exists the concept `Gateway` and `Routes`.
```
CoreyShupe: awk, maybe rephrase, also the plurality of `concept` suggests 1 concept.

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
...
+
+The return type is a `Completable` as defined by RxJava.
+This Completable will be subscribed to by ErisCasper.java.
+
+`Bot`s can be composed. Multiple implementations of `Bot` may be merged into a single `Bot`.
+
+
+## Event Stream
+
+The event stream is the events sent from Discord.
```
CoreyShupe: `consists of the events sent from discord`
A stream would consist of items, not be the array of events. imo it works better. Up for debate.

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
...
+This includes events such as message sends, new guild members, etc.
+
+The event stream uses the `Observable` type from RxJava to make the events available to bot developers.
+
+
+## Actions
+
+`Action`s are interactions with the Discord API that are expected to change its state,
+but for which there is no response expected.
+Examples include sending a message, or adding a role.
```
CoreyShupe: A comma isn't necessary after `message` when referring to 2 items of the same context. </nit>

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+## Repositories
+
+`Repository`s are the source of information from Discord.
+This could include, for example, information about guilds, channels, or users.
```
CoreyShupe: `This could include: information about guilds; channels; users; and many more forms of data which need to be cached`
May work better or some other variation.
Maybe `This could include, for example: information about guilds; channels; or users.`
I see it as better just because we use phrases in the list, `information about guilds`, which doesn't follow the normal convention of having a list of items. Up for debate though.

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
...
+
+`Repository` implmentations may perform caching, gateway requests, or HTTP requests
+as required.
+
+## Internal Concepts
+
+### Gateway
+
+Provide access to the (Discord Gateway)[https://discordapp.com/developers/docs/topics/gateway].
```
CoreyShupe: `provides`, the active form of the word

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
...
+This is a websocket provided by Discord used for tasks such as authentication and receiving events.
```
CoreyShupe: authentication
sending gateway commands
and receiving events

Has more than just the 2 uses described.

docs/core_api.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Routes
+
+Provides access to Discord's REST API.
```
CoreyShupe: We should reference this the same way we do the gateway.
https://discordapp.com/developers/docs/reference < this is the link I'd refer to.

CoreyShupe (changes requested):

docs/core_api.md
```diff
@@ -5,8 +5,8 @@ The core API of ErisCasper.java is designed to be minimal and powerful.
 It exposes the concepts of `Bot`s, the event stream (`Observable<Event>`), `Action`s, and `Repository`s as
 described below.
 
-There also exists the concept `Gateway` and `Routes`.
-These are lower level wrappers around the Discord API and considered internal
+At a lower level are `Gateway`s and `Routes`.
```
CoreyShupe: do we have multiple gateways?

CoreyShupe: nvm ignore this.

CoreyShupe (dismissed):

CoreyShupe (approved):

PR #172, opened 2018-07-02
Requests#execute refactor

ianagbip1oti:
Before this change we were building a request body using `data` and serializing it to json, and then ignoring it (the `buildRequest` function ignored the `body` parameter).

Now we avoid that and just use the one added in `route.newRequestBuilder(data)`

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/rest/Routes.java
+++ b/src/main/java/com/github/princesslana/eriscasper/rest/Routes.java
@@ -2,7 +2,6 @@ package com.github.princesslana.eriscasper.rest;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.princesslana.eriscasper.BotToken;
-import com.github.princesslana.eriscasper.rx.Maybes;
 import io.github.resilience4j.ratelimiter.RateLimiter;
 import io.github.resilience4j.ratelimiter.RateLimiterRegistry;
 import io.github.resilience4j.ratelimiter.operator.RateLimiterOperator;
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/princesslana/eriscasper/rest/Routes.java
```diff
@@ -47,15 +44,13 @@ public Routes(BotToken token, OkHttpClient client, ObjectMapper jackson) {
...
-    Function<Optional<RequestBody>, Request> buildRequest =
-        body ->
-            route.newRequestBuilder(data).header("Authorization", "Bot " + token.unwrap()).build();
-
-    return Maybes.fromNullable(data)
-        .map(d -> RequestBody.create(MEDIA_TYPE_JSON, jackson.writeValueAsString(data)))
-        .map(Optional::of)
-        .toSingle(Optional.empty())
-        .map(buildRequest)
+    Function<Request.Builder, Request.Builder> addAuthorizationHeader =
```
CoreyShupe, line 47: Should we consider making this a constant? or at least a private final field in the class. Also, the naming is kinda weird, verbs are mainly used as methods. A more conventional naming here would be something like `addAuthorizationHeaderFunction`. Just if we want to follow standard naming conventions. Not really big just a nit.

ianagbip1oti: We don't make local variables final in any other part of the code. There's no reason to define it outside the method because it's not used elsewhere. Duplicating the type (Function) in the name adds no extra information, especially as it has such limited scope.

CoreyShupe: gotcha, fair enough

CoreyShupe (approved):

PR #173, opened 2018-07-02
Simple conformity update for Data 0.6.3

CoreyShupe:
Inside of ErisCasper-Data 0.6.0 we:
- Moved all request classes to Data
- Moved all gateway commands to Data
- Remade Shard in Data
- Moved query string init into Data
+ Added in Nullable class to conform to Discord API
+ Made requests codegen so nullable and optional are applied properly.


CoreyShupe:
This needs a few testing fixes and work based on tests failing. A bit of a WIP, I'll tag and assign once it's ready for review.

ianagbip1oti:
Don't be afraid to break this up. e.g., do one PR that upgrades to 0.6.1 and has it compile/passes the tests. Then future PRs that change code, e.g., to use Shard from Data instead of Java.

CoreyShupe:
yeah I'm not too afraid about that, but simple class removal and method renaming is fine imo, especially since it's not much that needs to be discussed, it's a lot that's changing but most of it is just conformity to the new data introduced. Also, the setting up of passes and stuff require certain stuff (like the payload issue right now) I might revert the commit and redo a bit of the updating just to get it into a smaller PR then break it up more now that I think about it.

CoreyShupe:
Also @ianagbip1oti I need the Data issue fixed which I posted having to do with query strings before I can mess with this anymore. It's a vital change which kind of breaks everything since query strings aren't meant to have quotes. It was my fault in the end, just something which slipped through.

CoreyShupe:
Ready for comments. Heavily reduced and went through a lot of pom issues (just due to personal stuff with maven).

pom.xml
```diff
@@ -281,11 +281,11 @@
           <fail>true</fail>
           <failFast>false</failFast>
           <rules>
-            <dependencyConvergence></dependencyConvergence>
+            <dependencyConvergence />
```
ianagbip1oti: why?

src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
```diff
@@ -31,12 +32,13 @@
   private static final ChannelFunctionData<Guild> ADD_GUILD_CHANNELS_FUNCTION =
       ChannelFunctionData.of(
           (map, guild) -> {
-            if (guild.getChannels().isEmpty()) {
+            if (guild.getChannels().orElse(ImmutableList.of()).isEmpty()) {
```
ianagbip1oti, line 35: Is this the right approach?

If it's not present, don't we want to that that as 'no updates'?

src/test/java/com/github/princesslana/eriscasper/repository/event/TestChannelsFromEvents.java
```diff
@@ -34,7 +34,11 @@ public void getChannel_whenGuildCreated_shouldCacheChannel() {
     TestObserver<Channel> observer = new TestObserver<>();
 
     Guild guild = simpleCreateGuild();
-    Channel channel = guild.getChannels().get(0);
+    Channel channel =
+        guild
+            .getChannels()
+            .orElseThrow(() -> new IllegalStateException("Illegal state of list."))
```
ianagbip1oti: Since this is testing, doing a `get().get(0)` is ok. It'll throw an exception anyway.

ianagbip1oti (changes requested):

pom.xml
CoreyShupe: Cleaner, throwing warnings, kind of a code smell. Should I move that into a diff PR.

src/main/java/com/github/princesslana/eriscasper/repository/event/ChannelsFromEvents.java
CoreyShupe: If it's not present it will represent an empty list which has no updates. It works as you've explained.

pom.xml
ianagbip1oti: I disagree that it's cleaner or a code smell. What is giving warnings on this?

src/test/java/com/github/princesslana/eriscasper/repository/event/TestChannelsFromEvents.java
CoreyShupe: Alright, fair enough, just hate the yellow warnings that pop up for it. I'll change it.

pom.xml
CoreyShupe: It's telling me it's an empty code block. In xml you use `<x />` to represent an action rather than a block which it seems like this is. Using `<x> </x>` kind of says it has stuff inside of it. From what I know anyways, I might be wrong and I can change it back.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
```diff
@@ -81,7 +82,8 @@
       GuildFunctionData.of(
           (map, data) -> {
             Guild guild = map.get(data.getGuildId());
-            List<GuildMember> members = new ArrayList<>(guild.getMembers());
+            List<GuildMember> members =
+                new ArrayList<>(guild.getMembers().orElse(ImmutableList.of()));
```
ianagbip1oti, line 86: What you've said about orElse empty list makes sense in some places, but here isn't it going to delete members that are not in the list?

My understanding is that if it's not present, then we shouldn't be updating the member list.

pom.xml
ianagbip1oti: Talking about actions and code blocks in xml doesn't make sense to me. XML is just a data format.

An XML parser shouldn't treat `<x></x>` and `<x/>` any differently

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
CoreyShupe: If it's not present don't you think we should make it present and give the members of the guild a piece in the cache? If the members are assigned to the guild and the members aren't currently present why would we dispose of them? If it's not present and it's updating members at all we should know about the members. Also, I'm using the guild which is cached, they don't send a guild through along with the update. If the guild which is cached already doesn't have the members they don't exist to us.

ianagbip1oti: My issue is the kind of scenario where:
* We have guild members cached
* We receive an event where the list of members is null/absent
* We update the cache of guild members we have to empty list

pom.xml
CoreyShupe: See https://stackoverflow.com/questions/7231902/self-closing-tags-in-xml-files
It shouldn't treat it differently but from what I've seen it's `more correct` to use the self-closing tag in this instance.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
CoreyShupe: I don't see a case where this is relevant. I never use the given list of members as the list we use in the cache. I always use the list as a reference of what to update. If the list is empty it updates nothing.

pom.xml
CoreyShupe: We don't really comment it too much. Trying to find more accurate convention for xml self closing tags.

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
ianagbip1oti: Is this code here removing members based on what's in the cache and what's in the request? https://github.com/CoreyShupe/ErisCasper.java/blob/7118706dd4539a14f96befac1e30451512e7aa41/src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java#L101-L113

CoreyShupe: What this does is it takes the given list of items, takes the list turns it into a stream, and translates the active members into removable members if they appear in the container of removed snowflakes, then once it finds all of the snowflakes contained in the cache it removes the member version of them. It removes cached members based off information given in the request, not vice versa. If the list is empty nothing will happen because the container is empty. It's like querying an empty set if it has an item or not. It just returns no then moves on, therefore no members would actually be processed and the Stack would remain empty.

CoreyShupe: I think I worded that properly.

pom.xml
CoreyShupe: https://blogs.msdn.microsoft.com/ericwhite/2009/07/07/empty-elements-and-self-closing-tags/
Can also be seen here, seems like the popular opinion is that it provides a cleaner workspace. I can change it back if you'd like.

ianagbip1oti: I'm not sure how you get that from that post:
> Further, per the specification, “the empty-element tag SHOULD be used, and SHOULD only be used, for elements which are declared EMPTY”.  This means that it’s always safe to serialize an empty element as a self-closing element, but sometimes it’s not correct to serialize an empty element with a start and end tag.

and here's the style used in the pom for maven itself: https://github.com/apache/maven/blob/master/pom.xml#L529

src/main/java/com/github/princesslana/eriscasper/repository/event/GuildsFromEvents.java
ianagbip1oti: ok, cool

CoreyShupe: Also, once it removes the current set of members shown in the stack it adds all of the members back (as their updated version, just to stop members from showing up twice). I think there's actually a different error which occurs just due to object injection, which I'll fix.

CoreyShupe: So I actually worded that a bit wrong. End of the day, the members inside of the data get pushed into the cache, if empty nothing is changed, if there's stuff to remove it always just updates the cache. It never replaces the cache.

pom.xml
```diff
@@ -281,11 +281,15 @@
           <fail>true</fail>
           <failFast>false</failFast>
           <rules>
-            <dependencyConvergence></dependencyConvergence>
+            <dependencyConvergence>
+              <ignore />
```
ianagbip1oti: No, this is not what I meant with that link. I was showing they use tags that close themselves. `ignore` would be a configuration property to go with that particular plugin

CoreyShupe: Ahh alright, so we should use self-closing tags as well? Not sure what we want to do here.

ianagbip1oti: You can see the example for the enforcer plugin that we're using on this page: http://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html

CoreyShupe: Alright, so we should be using self-closing tags, as per what they're doing in their example.

CoreyShupe:
Also, the orElse() refers to a non-present set for the cached guild, not the data. If the guild has no data for members and we get updates for members we should definitely push them into the guild instead of leaving them not present.

ianagbip1oti (approved):

pom.xml
ianagbip1oti: sorry, I misread this whole change wrong from the start. I thought it was changing from `<x/>` to `<x></x>`.

`<x/>` is good

PR #176, opened 2018-07-03
Fix runtime error in BotContext#on

ianagbip1oti:


src/main/java/com/github/princesslana/eriscasper/BotContext.java
```diff
@@ -37,7 +37,9 @@ public BotContext(Observable<Event> events, Routes routes, RepositoryManager rep
 
   public <D, E extends Event & Wrapper<D>> Completable on(
       Class<E> evt, Function<D, Completable> f) {
-    return events.ofType(evt).map(Wrapper::unwrap).flatMapCompletable(f::apply);
+    // It would be nice to use a method reference here,
+    // but doing so causes an exception at runtime
+    return events.ofType(evt).map(e -> e.unwrap()).flatMapCompletable(f::apply);
```
CoreyShupe, line 42: Use `E::unwrap` not `Wrapper::unwrap`. Using E would source the generic not the wrapper. If it still produces a runtime it would be really weird since it must be the generic when pushed into from the class reference.

CoreyShupe (changes requested):

src/main/java/com/github/princesslana/eriscasper/BotContext.java
ianagbip1oti: I tried `E::unwrap` it also suffered the same issue. I have theories about why `::unwrap` approaches aren't working, but I haven't been able to recreate with a minimal example.

ianagbip1oti: Added a test for it that recreates the exception with `Wrapper::unwrap` or `E::unwrap`

CoreyShupe: interesting

CoreyShupe:
Can you post some of the test cases which it fails and throws the runtime exception along with the stack trace here just for reference. I think it might be useful, just in case we run into someone trying to do it again.

ianagbip1oti:
When using `Wrapper::unwrap` or `E::unwrap` it fails with an exception similar to this:

```
Exception in thread "main" java.lang.BootstrapMethodError: call site initialization exception
	at java.lang.invoke.CallSite.makeSite(CallSite.java:341)
	at java.lang.invoke.MethodHandleNatives.linkCallSiteImpl(MethodHandleNatives.java:307)
	at java.lang.invoke.MethodHandleNatives.linkCallSite(MethodHandleNatives.java:297)
	at com.github.princesslana.eriscasper.BotContext.on(BotContext.java:40)
	at com.github.princesslana.eriscasper.examples.WhoAreYouBot.apply(WhoAreYouBot.java:16)
	at com.github.princesslana.eriscasper.ErisCasper.run(ErisCasper.java:67)
	at com.github.princesslana.eriscasper.examples.WhoAreYouBot.main(WhoAreYouBot.java:35)
Caused by: java.lang.invoke.LambdaConversionException: Invalid receiver type interface com.github.princesslana.eriscasper.data.event.Event; not a subtype of implementation type interface com.github.princesslana.eriscasper.data.immutable.Wrapper
	at java.lang.invoke.AbstractValidatingLambdaMetafactory.validateMetafactoryArgs(AbstractValidatingLambdaMetafactory.java:233)
	at java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:303)
	at java.lang.invoke.CallSite.makeSite(CallSite.java:302)
	... 6 more
```

CoreyShupe:
I believe this is a JDK issue and has been fixed as per java 10, however, will not be supported in Java 8 or 9. The workaround is using a regular lambda expression instead of using a method reference. This is an issue with generics working with method references not being able to parse correctly or cast correctly. Just for reference as to why this doesn't work and what's happening with it.

CoreyShupe (approved):

ianagbip1oti:
If this is a JDK issue with a fix do you have a link to a bug report?

CoreyShupe:
https://bugs.openjdk.java.net/browse/JDK-8191655 covers it. I think it's been fixed since it's marked as resolved by the contributor assigned to it. It should be covered and fixed in Java10.

PR #177, opened 2018-07-03
Handle websocket closing

ianagbip1oti:
To be reviewed after #176.

This adds a test and ensures we throw an exception on any websocket closing or closed event. At the moment we have no way to recover from such an event and should not do nothing.

This is an improvement, but not yet optimal (e.g., on a Closing, Invalid Shard payload the exception is thrown and logged, we no longer send/receive any data, but the program does not stop).

Due to the behavior seen, some logic around stopping that is not working was also removed.

### DIFF
--- a/src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
+++ b/src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
@@ -90,17 +90,14 @@ public class Gateway {
   @SuppressWarnings("unchecked")
   public Observable<Event> connect(String url, BotToken token, Optional<Shard> shard) {
     Observable<RxWebSocketEvent> websocketEvents =
-        ws.connect(String.format("%s?v=%s&encoding=%s", url, VERSION, ENCODING));
-    websocketEvents
-        .ofType(RxWebSocketEvent.Closing.class)
-        .doOnNext(
-            e -> {
-              if (e.getCode() == 4004) {
-                e.getWebSocket().close(1002, "Invalid token.");
-                throw new ErisCasperFatalException("Failed to authenticate with discord servers.");
-              }
-            })
-        .share();
+        ws.connect(String.format("%s?v=%s&encoding=%s", url, VERSION, ENCODING))
+            .flatMap(
+                e ->
+                    e instanceof RxWebSocketEvent.Closing || e instanceof RxWebSocketEvent.Closed
+                        ? Observable.error(
+                            new ErisCasperFatalException("Websocket closed unexpectedly: " + e))
+                        : Observable.just(e));
+
     Observable<Payload> ps =
         websocketEvents
             .ofType(RxWebSocketEvent.StringMessage.class)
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -90,17 +90,14 @@ private void sequenceNumberSeen(Optional<SequenceNumber> seq) {
...
-                throw new ErisCasperFatalException("Failed to authenticate with discord servers.");
-              }
-            })
-        .share();
+        ws.connect(String.format("%s?v=%s&encoding=%s", url, VERSION, ENCODING))
+            .flatMap(
+                e ->
+                    e instanceof RxWebSocketEvent.Closing || e instanceof RxWebSocketEvent.Closed
+                        ? Observable.error(
+                            new ErisCasperFatalException("Websocket closed unexpectedly: " + e))
```
CoreyShupe, line 98: How do we know it's unexpected? We should check the reason, then emit it with that. I just don't like the message sent with this. I put it in but thinking about it more it's kinda bad.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -90,17 +90,14 @@ private void sequenceNumberSeen(Optional<SequenceNumber> seq) {
...
   public Observable<Event> connect(String url, BotToken token, Optional<Shard> shard) {
     Observable<RxWebSocketEvent> websocketEvents =
-        ws.connect(String.format("%s?v=%s&encoding=%s", url, VERSION, ENCODING));
-    websocketEvents
-        .ofType(RxWebSocketEvent.Closing.class)
-        .doOnNext(
-            e -> {
-              if (e.getCode() == 4004) {
-                e.getWebSocket().close(1002, "Invalid token.");
-                throw new ErisCasperFatalException("Failed to authenticate with discord servers.");
```
CoreyShupe, line 100: This is a pretty specific code for Invalid tokens. Are we sure we shouldn't be handling it separately since we know what the code translates to so we can notify the user of the issue?

CoreyShupe (changes requested):

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
ianagbip1oti: The approach I've taken for now is to consider it unexpected always.

As mentioned, this isn't optimal, we can add specific handling for other cases (we could probably use something more specific than FatalException for all of them also).

There are details provided in the message, as the reason that discord sends is included there.

CoreyShupe: Alright, in future PRs we should build up this to include everything needed. This can be used as the structure though. Works for me.

ianagbip1oti: Right, at the moment there isn't a clean shutdown when this happens, in that the program doesn't exit.

Once it's understand the best way to have it exit correctly, we can visit ways of ending it in a way that gives the best message to the bot author.

CoreyShupe (approved):

PR #178, opened 2018-07-03
RxWebSocket refinement

ianagbip1oti:
When we receive onClosing it means a graceful closure of the websocket has been requested, so we should close the websocket.

On the closed event we should complete the Observable.

Added tests for this functionality, and split tests into more focused smaller tests.


src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -51,11 +50,13 @@ private Listener(ObservableEmitter<RxWebSocketEvent> em) {
     @Override
     public void onClosed(WebSocket ws, int code, String reason) {
       em.onNext(ClosedTuple.of(ws, code, reason));
+      em.onComplete();
```
CoreyShupe, line 56: Are we sure there's no `complete()` method? onComplete seems more like a runnable function rather than a method which actually closes the observable.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
```diff
@@ -51,11 +50,13 @@ private Listener(ObservableEmitter<RxWebSocketEvent> em) {
...
     }
 
     @Override
     public void onClosing(WebSocket ws, int code, String reason) {
       em.onNext(ClosingTuple.of(ws, code, reason));
+      ws.close(0, "Close.");
```
CoreyShupe: We shouldn't use `Close.` as the reason. Try something having to do with `reason` or some other way of finding out what it should be. ws.close() takes websocket closing params with a reason. I don't see `close.` as a good reason for this in any way. https://docs.oracle.com/javaee/7/api/javax/websocket/CloseReason.CloseCodes.html you could use this as a guide for codes. `0` I don't think is a valid one, especially since we don't know if it's graceful.

CoreyShupe (changes requested):

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
ianagbip1oti: It's used to signal completion: http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Emitter.html#onComplete--

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
ianagbip1oti: Good call on the status code. I'll update that.

We do know that it's graceful, as per the closing state documented here: https://square.github.io/okhttp/3.x/okhttp/okhttp3/WebSocket.html

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: Gotcha.

src/main/java/com/github/princesslana/eriscasper/rx/websocket/RxWebSocket.java
CoreyShupe: Even if it is graceful we should pass `reason` through for the reason on ws.close(). I think it's useful for transmitting it to the user rather than letting it dissipate. Also useful for telling the other side why we're closing. `Close.` isn't very informative.

CoreyShupe: Also, the closing statement there doesn't mean the close was graceful, it just means it's closing the socket gracefully. We don't know how graceful the endpoint was in talking back to us. So the reason might be good to pass through. Even if we just send a graceful code as per `1001`.

ianagbip1oti: At the network level, it is graceful. The other end has told us they're closing the connection, we respond by closing it. There's no network failures or cutoffs or other action we can take.

At the semantic/user of the bot level, maybe it's not, but that's already been communicated back up the stack by emitting the ClosingTuple event.

CoreyShupe: Alright, fair enough, maybe use something like this:
`ws.close(1001, "Shutting down.");`
imo `Close.` isn't good enough in the circumstance, it feels insufficient for what it's asking for.
I don't think `Close.` is a `reason`, idk more of a grammatical issue in a sense, nitpicking ofc.

ianagbip1oti: Have updated status code to 1000, 'cos for me this is just the underlying transport behaving like it should, nothing abnormal. I've removed the message since it's optional and adds no info.

CoreyShupe: Fair enough. I think this is a good decision.

CoreyShupe (approved):

CoreyShupe:
We might want to merge this after PR #177

PR #180, opened 2018-07-04
Updated shard usage to Data shard payload instead of .Java shard.

CoreyShupe:
Simple shard update. Conformity to new version of Data.


src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -77,11 +78,42 @@ public static ErisCasper create() {
...
+        BotToken.of(token), Optional.of(ShardPayload.of(shardNumber, shardTotal)));
+  }
+
+  private static void assertShard(long shard, long total) {
+    Preconditions.checkArgument(shard >= 0, "Shard number must be greater than or equal to 0.");
+    Preconditions.checkArgument(total >= 1, "Shard total must be greater than or equal to 1.");
+    Preconditions.checkState(shard < total, "Shard number must be less than the shard total.");
+  }
+
+  public static Optional<ShardPayload> shardFromConfig(Properlty config) {
```
ianagbip1oti: If we have a number of methods where we want to put `shard` in the title, this suggests it should be on the Shard class, but given that we generate it, we can't do that. The convention used is, to create a `Shards` class to add this methods. We can then have `Shards.check` and `Shards.fromConfig` methods.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -77,11 +78,42 @@ public static ErisCasper create() {
...
+              .map(
+                  s ->
+                      total
+                          .map(
+                              t -> {
+                                assertShard(s, t);
+                                return ShardPayload.of(s, t);
+                              })
+                          .orElseThrow(() -> exception))
+              .orElseThrow(() -> exception));
```
ianagbip1oti: Feels like there should be a better way

ianagbip1oti (changes requested):
Has it been tested that the bots start up and identify successfully when no shard is provided?

The upgrade to 0.6.3 of data may have fixed this.

CoreyShupe:
I'm fairly certain we have tests which use no shards on the connection. Either way the Optional nature of ShardPayload allows it to have nothing. The filtering should have picked it up and fixed it, if it was an issue before.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: I could do presence checks instead of mapping checks. It would do the same thing virtually. It all should be chained though and should work just as fast. It just looks kind of wonky.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: Alright, I'll move it.

ianagbip1oti:
It was definitely an issue before. We were sending `shard: null` on the identify request, which discord reject as invalid. Our existing tests did not detect this.

It should be fixed with the changes to Data, but it would be good to confirm

CoreyShupe:
I think it's confirmed through the filtering tests we have in Data. Since it comes empty it's been proven by our previous versions. There's no need to confirm filtering here. The issue came from the lack of filtering not from the issue with sharding. I don't think this is the right place to test it.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe:
```java
  public static Optional<ShardPayload> fromConfig(Properlty config) {
    Optional<Integer> shard = config.getInt("ec.shard.id");
    Optional<Integer> total = config.getInt("ec.shard.total");
    if (shard.isPresent() ^ total.isPresent()) {
      throw new ErisCasperFatalException(
          "Failed to resolve both sharding values when only one was provided.");
    }
    return shard.map(
        s -> {
          long t = total.get();
          check(s, t);
          return ShardPayload.of(s.longValue(), t);
        });
  }
```
This would be a good alternate route if you want to take it.

CoreyShupe: Or even
```java
  public static Optional<ShardPayload> fromConfig(Properlty config) {
    Optional<Long> shard = config.getLong("ec.shard.id");
    Optional<Long> total = config.getLong("ec.shard.total");
    if (shard.isPresent() ^ total.isPresent()) {
      throw new ErisCasperFatalException(
          "Failed to resolve both sharding values when only one was provided.");
    }
    return shard.map(
        s -> {
          long t = total.get();
          check(s, t);
          return ShardPayload.of(s, t);
        });
  }
```
to stop senseless long casting

ianagbip1oti:
Have you started up a bot and confirmed it can identify?

I'm not necessarily after an automated check here. I agree it should already be fixed, I'm just asking for a manual one time check. For example, there's a chance it doesn't go via Data, but instead has an ObjectMapper configured somewhere else.

CoreyShupe:
I can write up a test which verifies it's empty and not null if you'd like. I don't think a manual test is necessary since we know the problem and know when it's fixed. (also I can't connect to discord.gg or discordapp.com so I can't really test it on my end).

ianagbip1oti:
Do we know it's fixed? AFAIK no one has run a bot since what we think was the fix was merged.

CoreyShupe:
You'd need to be the one to test it, I just can't connect to the API endpoints due to school restrictions.

schnapster:
Master branch now works for me where it was previously broken.

CoreyShupe:
Thank you @napstr now we have a manual check for assurance. <3 much love

src/main/java/com/github/princesslana/eriscasper/Shards.java
```diff
@@ -0,0 +1,36 @@
...
+  public static Optional<ShardPayload> fromConfig(Properlty config) {
+    Optional<Long> shard = config.getLong("ec.shard.id");
+    Optional<Long> total = config.getLong("ec.shard.total");
+    if (shard.isPresent() ^ total.isPresent()) {
+      throw new ErisCasperFatalException(
+          "Failed to resolve both sharding values when only one was provided.");
+    }
+    return shard.map(
+        s -> {
+          long t = total.get();
```
ianagbip1oti, line 21: I reckon if we're going to use get (which is ok, once you start nesting map inside map inside map things get messy unfortunately), I think we should use it for both. Since one is from a map, and one from get, took me way longer than it should to work out where `s` was coming from :D

ianagbip1oti (approved):

src/main/java/com/github/princesslana/eriscasper/Shards.java
CoreyShupe: Oh haha, is that an issue you want me to work with or change? Or is it just kinda wonky haha, I can rename variables and such if needed.

CoreyShupe: Also, using map for both doesn't really make sense, only if we used get for both, but mapping keeps it as an optional, which is what we're looking for.

ianagbip1oti: I approved it anyway. It feels like there should be a better way, but I can't think of it 🤷‍♀️

CoreyShupe:
This should be merged then. Next action rewrite.

PR #182, opened 2018-07-04
Rewrite of Action API

CoreyShupe:


src/main/java/com/github/princesslana/eriscasper/action/ActionContext.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.github.princesslana.eriscasper.data.util.Nullable;
+import com.github.princesslana.eriscasper.gateway.Gateway;
+import com.github.princesslana.eriscasper.rest.Routes;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface ActionContext {
+
+  Nullable<Routes> getRoutes();
```
ianagbip1oti: In what cases do we expect this to be null?

src/main/java/com/github/princesslana/eriscasper/action/VoidRouteAction.java
```diff
@@ -0,0 +1,23 @@
+package com.github.princesslana.eriscasper.action;
+
+import com.github.princesslana.eriscasper.rest.Route;
+import io.reactivex.Completable;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface VoidRouteAction<O> extends Action {
```
ianagbip1oti: How much do we gain with a separate VoidRouteAction, over just using RouteAction<Void, T>?

src/main/java/com/github/princesslana/eriscasper/action/GatewayAction.java
```diff
@@ -0,0 +1,26 @@
+package com.github.princesslana.eriscasper.action;
+
+import com.github.princesslana.eriscasper.gateway.OpCode;
+import io.reactivex.Completable;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface GatewayAction<I> extends Action {
```
ianagbip1oti: I'd actually prefer these to be abstract class, since execute is real functionality vs something computed from other fields.

I think semantically it's cleaner to keep real functionality out of interfaces

src/main/java/com/github/princesslana/eriscasper/action/Action.java
```diff
@@ -1,32 +1,8 @@
...
-/** An Action is a {@link Route} with the data required to execute it. */
-@Value.Immutable
-public interface Action<I, O> {
-  /**
-   * Get the {@link Route} that this Action will call.
-   *
-   * @return the Route this Action will call
-   */
-  Route<I, O> getRoute();
+public interface Action {
```
ianagbip1oti: Is there value in having this extend Function<ActionContext, Completable>

Probably reactivex Function rather than java's function.

ianagbip1oti (changes requested):

src/main/java/com/github/princesslana/eriscasper/action/Action.java
CoreyShupe: We could do that, good call.

src/main/java/com/github/princesslana/eriscasper/action/GatewayAction.java
CoreyShupe: I can make the abstract classes, yeah, sounds like a good idea

src/main/java/com/github/princesslana/eriscasper/action/VoidRouteAction.java
CoreyShupe: `execute(Route<Void, X>)` is a diff method than `execute(Route<X, Y>, X)`
With the functionality we're trying to produce I can't check generic attributes without heavy reflection. Using `RouteAction<Void, X>` wouldn't be the proper way to pass through the method. However, if using null as a piece of data to represent void works then we can use that, but I'm not sure if that works, if it does I'll remove it.

src/main/java/com/github/princesslana/eriscasper/action/ActionContext.java
CoreyShupe: Test cases use null routes.

src/main/java/com/github/princesslana/eriscasper/action/VoidRouteAction.java
ianagbip1oti: Using `null` should work.

src/main/java/com/github/princesslana/eriscasper/action/ActionContext.java
ianagbip1oti: While it's true that easily testable code is often well designed code, testing should never be a driver of design decisions.

src/main/java/com/github/princesslana/eriscasper/action/VoidRouteAction.java
CoreyShupe: Then I'll remove it and if something breaks I'll add it back.

src/main/java/com/github/princesslana/eriscasper/action/ActionContext.java
CoreyShupe: Alright, I may keep the gateway null since it makes sense that you can provide routes without a gateway to process commands through. I'll need to redesign some of the tests we have though, due to the removal of the `Nullable<Routes>`. May take a bit of work just since in test cases for certain bot context methods we aren't able to produce a viable `Routes` but then again we can make a FakeRoutes class. We could also make a FakeGateway pass through as well, if we want to remove the nullable attribute from that as well (which might be a good idea)

ianagbip1oti: You can easily make a mock Routes or Gateway for testing.

The only place we create an `ActionContext`, is from `BotContext`. There are no cases we have at the moment that would require us to create an `ActionContext` without a `Gateway`.

CoreyShupe: Would just using Mockito provide enough mocking for the Routes and Gateway?

ianagbip1oti: For testing, yes.

CoreyShupe:
Inside of `Actions`, should we be returning `Action` for everything rather than their complete parts? So if it's a `RouteAction`, just return `Action`. Just so the user isn't exposed to things like `RouteAction` and `GatewayAction`.

src/main/java/com/github/princesslana/eriscasper/BotContext.java
```diff
@@ -51,21 +55,20 @@ public Completable doNothing() {
    *
    * <p>(As opposed to a Query)
    */
-  public <I, O> Completable execute(Action<I, O> action) {
-    LOG.debug("Executing {}...", action);
-    return execute(action.getRoute(), action.getData()).toCompletable();
+  public Completable execute(Action action) {
+    return Single.just(actionContext).flatMapCompletable(action);
```
ianagbip1oti: Why not just `action.apply(actionContext)`?

CoreyShupe: `io.reactivex.function.Function` produces an `Exception` exception on `Function#apply(X)`. The way to handle that error respectively in a Completable context is by passing it through a `Single`.

ianagbip1oti:
I'm 50/50 on whether we should return `Action` or `RouteAction`/`GatewayAction`.

We do want the user to do things like `Action a = Actions.sendMessage("blah")`, rather than use `RouteAction` or `GatewayAction`. But that can do that whichever we return, and `RouteAction`/`GatewayAction` more accurately reflect the return type of the methods.

I'm fine with either way at this point in time.

CoreyShupe:
I'll just leave it, I'm not scared of exposing things like the route or data, I will be doing it for gateway commands so the OpCode isn't exposed to the user.

src/main/java/com/github/princesslana/eriscasper/BotContext.java
ianagbip1oti: As an option here, for `Bot`, we extend `Function`, but define it without an exception: https://github.com/princesslana/ErisCasper.java/blob/master/src/main/java/com/github/princesslana/eriscasper/Bot.java#L19

CoreyShupe: I think keeping in the checked exception is good, I think it produces the exception in a clean manner. Especially since we're providing a `Completable` context. We want to cleanly handle all exceptions thrown by `Routes` or the `Gateway` inside of the stream. I think it's good how it is. If it's a problem or hurts efficiency I can change it, but I think it's the cleanest thing to do.

CoreyShupe: Especially for making it an exposed API, so the user can throw exceptions inside the stream without being worried about them being handled poorly.

ianagbip1oti: My thinking is this:
* The only reason we use reactivex `Function` instead of Java `Function` is because it works nicer with the reactivex streams.
* In general `throws Exception` is bad, and along with  https://rongi.github.io/kotlin-blog/rxjava/rx/2017/08/01/error-handling-in-rxjava.html, I don't think we want our Actions throwing an exception.

CoreyShupe: That's a fair point. I'll change it.

CoreyShupe:
@ianagbip1oti any updates for this, or comments?

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -42,14 +42,16 @@
...
+
   private ErisCasper(BotToken token, Optional<ShardPayload> shard) {
     this.token = token;
     this.shard = shard;
     routes = new Routes(token, httpClient, jackson);
   }
 
   private Observable<Event> getEvents() {
-    Gateway gateway = Gateway.create(httpClient, payloads);
+    this.gateway = Gateway.create(httpClient, payloads);
```
ianagbip1oti: Why can't we create this in the constructor?

src/main/java/com/github/princesslana/eriscasper/action/GatewayAction.java
```diff
@@ -0,0 +1,23 @@
...
+import com.github.princesslana.eriscasper.gateway.OpCode;
+import io.reactivex.Completable;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class GatewayAction<I> implements Action {
+
+  public abstract OpCode getCode();
+
+  // There is no gateway action which takes no data
```
ianagbip1oti: I feel this comment is redundant. It's implied by not using Optional. </nit>

CoreyShupe: Gotcha, I was just defining why it was different from `RouteAction`

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: 2 different methods, we could merge the methods but that would break the separation we want to maintain.

CoreyShupe: Oh wait, it's merged into 1, didn't notice.

CoreyShupe: Yeah, they come into 2 separate methods.
https://hastebin.com/rorokiluxu.java
If we merged them it would ruin the flow of the bot.

CoreyShupe: We could hold the botcontext as a field in the ErisCasper instance but it might not be worth it to do that over the gateway.

CoreyShupe: If you're referring to the ErisCasper constructor we could do that and finalize the gateway variable. Didn't really understand what you were going at until just now.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -36,20 +36,21 @@
...
   private final Routes routes;
   private final Optional<ShardPayload> shard;
 
+  private final Gateway gateway;
+
   private ErisCasper(BotToken token, Optional<ShardPayload> shard) {
     this.token = token;
     this.shard = shard;
-    routes = new Routes(token, httpClient, jackson);
+    ObjectMapper jackson = Jackson.newObjectMapper();
```
ianagbip1oti, line 48: Just a note for later that we need to remove the need to do this. Everything should go through the Data.xxx methods.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
ianagbip1oti: Yeah, in the ErisCasper constructor. With final and then we know it's always not null too 👍

ianagbip1oti (approved):

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
CoreyShupe: The only reason this is there is because it was there originally, I'm going to go into Routes and Payloads and replace the methods later, but for now, it can stay.

PR #183, opened 2018-07-05
Clean shutdown on websocket closing

ianagbip1oti:
The changes here:
* Remove the exception on websocket closing. I tend to agree with this: https://rongi.github.io/kotlin-blog/rxjava/rx/2017/08/01/error-handling-in-rxjava.html, in that things that aren't bugs throw Exceptions works out badly in RxJava. So *for now* we just log at a warning level.
* Add `cleanupDaemonThreads` to the maven build - otherwise we have to wait for OkHttp daemon threads to shutdown (there is no force shutdown for the daemon threads)
* Refactor ErisCasper.run so that the bot is in the Event stream - this means the event stream completing means completion in the run method.
* Add cleanup of RxJava and OkHttp (what we can) when the bot completes

The end result of this is an immediate shutdown if a bad bot token is passed, with out a 1k long stacktrace.

Address #71


src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -87,19 +87,19 @@ private void sequenceNumberSeen(Optional<SequenceNumber> seq) {
     seq.ifPresent(sid -> lastSeenSequenceNumber = Optional.of(sid));
   }
 
-  @SuppressWarnings("unchecked")
+  private void warnOnClosing(RxWebSocketEvent evt) {
+    if (evt instanceof RxWebSocketEvent.Closing || evt instanceof RxWebSocketEvent.Closed) {
+      LOG.warn("Websocket closing: {}", evt);
```
schnapster: I have found information like the endpoint address (probably not relevant for the main websocket) and the shard id logged together with close events (code and reason) very helpful in some situations in the past (discord devs usually also require this information, when reporting rare:tm: bugs). Not sure how much of that information is available in the current Gateway implementation here; if it's available, it's worth adding imo.

CoreyShupe: I think debugging and finding ways to show information for bugs should be done later when we're producing 1.0.0, not now, I don't think it's necessary to provide general debugging for users when the lib isn't even in a stable version yet.

CoreyShupe: There's also probably going to be a lot of work done on the websocket still. Not worth providing user driven stuff for items we're just going to change. Me and Lana personally should have enough information when debugging or have the ability to get the information from a debugging process and reproduction of bugs when testing. Also our sharding isn't quite finished yet or in a production stage. It might be useful later on but there's a big TODO stamp in this section imo. It's not quite in a stage which provides a bunch of information (probably will be done in the documentation stages as well)

ianagbip1oti: The reason and code where already there. Have updated to include url and shard. Logged twice 'cos we log on closing and closed events.

```
[OkHttp https://gateway.discord.gg/...] WARN com.github.princesslana.eriscasper.gateway.Gateway - Websocket closing: url=wss://gateway.discord.gg, shard=Optional.empty, event=Closing{webSocket=okhttp3.internal.ws.RealWebSocket@10d5381, code=4004, reason=Authentication failed.}
[OkHttp WebSocket https://gateway.discord.gg/...] WARN com.github.princesslana.eriscasper.gateway.Gateway - Websocket closing: url=wss://gateway.discord.gg, shard=Optional.empty, event=Closed{webSocket=okhttp3.internal.ws.RealWebSocket@10d5381, code=4004, reason=Authentication failed.}
```

CoreyShupe: That too ^

CoreyShupe: I think there's still more we could do with providing all the information possible to everywhere in every case of exception. I'll look around of updates for information being passed. But imo that would be done when we start expanding the exception list for ErisCasper to start capturing specific cases.

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -60,13 +60,21 @@ private ErisCasper(BotToken token, Optional<ShardPayload> shard) {
...
-        .blockingAwait();
+    try {
+      getEvents()
+          .compose(
+              evts ->
+                  bot.apply(new BotContext(evts, routes, RepositoryManager.create(evts)))
+                      .toObservable())
+          .doOnError(t -> LOG.warn("Exception thrown by Bot", t))
+          .blockingSubscribe();
+    } finally {
```
CoreyShupe, line 71: Should we catch and display `ErisCasperFatalException`s here?

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -87,19 +88,21 @@ private void sequenceNumberSeen(Optional<SequenceNumber> seq) {
...
-                        ? Observable.error(
-                            new ErisCasperFatalException("Websocket closed unexpectedly: " + e))
-                        : Observable.just(e));
+    CompositeDisposable disposables = new CompositeDisposable();
 
     Observable<Payload> ps =
-        websocketEvents
+        ws.connect(String.format("%s?v=%s&encoding=%s", url, VERSION, ENCODING))
+            .doOnNext(warnOnClosing(url, shard))
+            .doFinally(disposables::dispose)
```
CoreyShupe, line 105: Are we sure this captures and produces the hearbeat/identify etc... as expected? Do we have some manual testing to back up this change?

src/test/java/com/github/princesslana/eriscasper/gateway/TestGateway.java
```diff
@@ -53,6 +51,8 @@ public void mockito() {
 
   @BeforeMethod
   public void subject() {
+    wsEvents = PublishSubject.create();
```
CoreyShupe, line 54: is there a reason we moved this?

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
ianagbip1oti: They would be displayed in the `doOnError`, or we just let them bubble up, where if the bot dev isn't catching them they'll display when the program exits anyway.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
ianagbip1oti: It produces the capture/heartbeat, I can startup PingBot, it responds, I can see the heartbeat being sent periodically.

src/test/java/com/github/princesslana/eriscasper/gateway/TestGateway.java
ianagbip1oti: Yes, we want the subject to be reset for each test. Otherwise `ws.onComplete` completes it for all other tests.

CoreyShupe:
Alright, going to go more in depth looking into it tomorrow, a bit tired tonight, just got back from fireworks.

pom.xml
```diff
@@ -301,6 +301,7 @@
           </dependency>
         </dependencies>
         <configuration>
+          <cleanupDaemonThreads>false</cleanupDaemonThreads>
```
CoreyShupe, line 304: why don't we want these cleaned up?

src/main/java/com/github/princesslana/eriscasper/ErisCasper.java
```diff
@@ -60,13 +60,21 @@ private ErisCasper(BotToken token, Optional<ShardPayload> shard) {
...
+      Schedulers.shutdown();
+      httpClient.dispatcher().executorService().shutdown();
+      httpClient.connectionPool().evictAll();
+    }
+
+    LOG.trace("ErisCasper.run done.");
```
CoreyShupe: should we change this to a method reference? E.g `Log.trace("ErisCasper#run(Bot) done.");`

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -108,26 +111,24 @@ private void sequenceNumberSeen(Optional<SequenceNumber> seq) {
...
-            .flatMapCompletable(
-                p -> isResumable() ? resume(ws, token) : identify(ws, token, shard));
+            .flatMapCompletable(p -> heartbeat(ws, p))
+            .subscribe());
 
-    Observable<Event> events = ps.flatMapMaybe(payloads::toEvent).share();
+    disposables.add(
+        ps.filter(Payload.isOp(OpCode.HELLO))
+            .flatMapCompletable(p -> isResumable() ? resume(ws, token) : identify(ws, token, shard))
+            .subscribe());
```
CoreyShupe, line 122: Could we somehow merge the 2 filtering methods? They're both filtering out the `OpCode.HELLO`

src/test/java/com/github/princesslana/eriscasper/gateway/TestGateway.java
```diff
@@ -1,12 +1,12 @@
 package com.github.princesslana.eriscasper.gateway;
 
+import static org.mockito.BDDMockito.any;
```
CoreyShupe, line 3: Why are we importing this in a static manner? We use it once, I don't think it deserves its own import.

CoreyShupe (changes requested):

pom.xml
ianagbip1oti: Because the okhttp daemon threads take ~ 15 seconds to cleanup. With this set to false maven completes and the daemon threads are cleaned up faster.

src/test/java/com/github/princesslana/eriscasper/gateway/TestGateway.java
ianagbip1oti: For me importing it is consistent with how we use `BDDMockito`. Having `BCCMockito` in the fluent API ruins its flow

`given(mockRxWebSocket.send(any()))`
`given(mockRxWebSocket.send(BDDMockito.any()))`

FWIW, I'm usually against static imports, but BDDMockito's API gets very noisy without them.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
ianagbip1oti: Maybe, but then we'd have to do one of the following:
* Add another share, so that we can add multiple flows for `ps.filter(Payload.isOp(OpCode.HELLO)`
* Add a method like `onHello(ps)`, which doesn't really gain us much readability, but it does add another method.

CoreyShupe: Could we not do something like:
```java
    disposables.add(
        ps.filter(Payload.isOp(OpCode.HELLO))
            .flatMapCompletable(p -> isResumable() ? resume(ws, token).merge(heartbeat(ws, p)) : identify(ws, token, shard).merge(heartbeat(ws, p)))
            .subscribe());
```
or even
```java
    disposables.add(
        ps.filter(Payload.isOp(OpCode.HELLO))
            .flatMapCompletable(p ->{ 
              Completable hearbeat = hearbeat(ws, p);
              return isResumable() ? resume(ws, token).merge(heartbeat) : identify(ws, token, shard).merge(heartbeat)
            })
            .subscribe());
```
or
```java
    disposables.add(
        ps.filter(Payload.isOp(OpCode.HELLO))
            .flatMapCompletable(p ->{ 
              Completable hearbeat = hearbeat(ws, p);
              return hearbeat.merge(isResumable() ? resume(ws, token) : identify(ws, token, shard));
            })
            .subscribe());
```

src/test/java/com/github/princesslana/eriscasper/gateway/TestGateway.java
CoreyShupe: Gotcha, it's more nitpicking and if it's less noisy then it's worthwhile.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
ianagbip1oti: one and two both keep some duplication in the form of the heartbeat merge being in multiple places. The last one removes the duplication, but I'm not sure how much more readable it is.

For now, I'd like to keep them separate because it's not too much duplication, and we may end up wanting more logic on the heartbeat here (for example, if we received another Hello now I don't think we stop the first heartbeat, meaning we'd be having two heartbeats being sent)

CoreyShupe: Alright, it's not a huge issue, just nitpicking. We can keep them separate especially if you have more plans for logic.

CoreyShupe (approved):

PR #184, opened 2018-07-12
Removed all .Java gateway commands and moved into -Data commands.

### LINKED ISSUE #122: Support all Gateway Commands

CoreyShupe:
Changed usage of gateway commands such as `Identify` and `Resume` to use the ErisCasper-Data versions.
Fixes #122


src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -161,10 +161,10 @@ private Completable resume(RxWebSocket ws, BotToken token) {
         lastSeenSequenceNumber.isPresent(), "Can not resume without a sequence number");
 
     return Single.just(
-            ImmutableResume.builder()
-                .token(token)
-                .sessionId(sessionId.get())
-                .seq(lastSeenSequenceNumber.get())
+            ImmutableResumePayload.builder()
+                .token(token.unwrap())
+                .sessionId(sessionId.get().unwrap())
```
ianagbip1oti: I realize get was used before, but what are we calling `get` on here? A nullable or optional? Can we do better?

CoreyShupe: Optional, we can do better, I replaced it with a much better system in my opinion with better errors thrown.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -156,15 +155,23 @@ private Completable heartbeat(RxWebSocket ws, Payload hello) {
...
+                            () ->
+                                new IllegalStateException("Can not resume without a session id.")))
+                .seq(
+                    lastSeenSequenceNumber
+                        .map(SequenceNumber::unwrap)
+                        .map(Integer::longValue)
+                        .orElseThrow(
+                            () ->
+                                new IllegalStateException(
+                                    "Can not resume without a sequence number.")))
```
ianagbip1oti: Do we know why these are marked as optional in the api if we can't resume with them?

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
```diff
@@ -156,15 +155,23 @@ private Completable heartbeat(RxWebSocket ws, Payload hello) {
...
+                .sessionId(
+                    sessionId
+                        .map(SessionId::unwrap)
+                        .orElseThrow(
+                            () ->
+                                new IllegalStateException("Can not resume without a session id.")))
+                .seq(
+                    lastSeenSequenceNumber
+                        .map(SequenceNumber::unwrap)
+                        .map(Integer::longValue)
```
ianagbip1oti: If this is meant to be a long, then SequenceNumber should probably wrap a long instead of an integer

CoreyShupe: Agreed, I'll look through and change them as needed.

src/main/java/com/github/princesslana/eriscasper/gateway/Gateway.java
CoreyShupe: Not sure, I'll find out and change them if necessary. I think it's just to stop resuming if we don't have the correct stuff. I think there's a way to have it try to resume with null values and having null values is a pain. We might want to change them to `Nullable<X>`

CoreyShupe: So `SequenceNumber` is optional coming from a `Payload`, however inside of the `Gateway` it should be Nullable. `SessionId` should be `Nullable` regardless.

ianagbip1oti: Why should it be nullable instead of optional?

CoreyShupe: SequenceNumber should be Optional, SessionId should be nullable.
SessionIds are this or null, they aren't optional, SequenceNumbers have an issue in the heartbeat parsing into a payload where they need to be optional for them to actually be converted properly.

ianagbip1oti: Source on session ids can be null? vs the optional case where we either know it or we don't?

CoreyShupe: So SessionIds start off not there, they start off `null`, then once we start talking to the server, if they talk to us properly, then we have the SessionId, if not it remains `null`, however, we cannot resume without a SessionId meaning in all cases it's not an optional item. SequenceNumbers have a property in payloads where they can be optional and don't have to be there when talking, only in heartbeats and certain other payloads. Hence why SequenceNumbers are optional and SessionIds are not. Honestly, it's just the issue of what the payload is accepting and how we translate.

ianagbip1oti: Where do that start off null? If they are not there, isn't that just `Optional.empty()`?

CoreyShupe: We start off null before we start talking to discord, it would be `Optional.empty()` except I have trouble making it `Optional` at all since in all cases where the SessionId is present they aren't optional, they're required. I though have issue making any field `null` since it's just bad practice and it's harder to handle and deal with, having it `Nullable` gives us access to nice methods we can use for cleaner exceptions and better mapping of the value.

ianagbip1oti: So when we need to send a SessionId we are allowed to send null, but we're not allowed to omit it? If so it does make sense for it to be Nullable, but where in the API is that?

CoreyShupe: We can't omit it and we can't have it `null`. That's why I throw an exception when it is `null`. afaik there's no place other than `Resume` which uses it.

ianagbip1oti: Ok, so I'm not clear what's stopping it from being Optional, or what we gain from it being Nullable. We only introduced Nullable because the discord api forces us to. I don't think we want to start using Nullable outside of cases where the discord api forces us to.

CoreyShupe: `Optional` imo just isn't a good identifier for the field. We could just make the field regular and have it as `null` and setup preconditions, but it feels like that's a dirty way of doing it.

CoreyShupe: It just doesn't make sense to be `Optional` since it isn't `Optional`. We require it anywhere we use it. The reason I'm so hesitant on making it `Optional` is because it defines the field horribly in ways which just aren't true.

ianagbip1oti: Why is it not good?

Discord never tells us it's null.

So, we have either been given a value, or we don't have a value. Isn't this exactly what Optional represents with empty/present?

I know that we can represent it with null/value, or with Nullable.null/present, but Optional was introduced to overcome the null/value case, and we only introduced Nullable for Discord API reasons. We didn't introduce Nullable as a 'better null'.

CoreyShupe: Yeah, I get where you're coming from and I see your point. I just don't like maintaining a field as Optional which isn't Optional anywhere we use it. I can change it back to `Optional<X>` though, it's not a big issue.

ianagbip1oti: It's not Nullable either :)

I think you're getting hung up on the name (e.g., in other languages it's called Maybe). For me it's intended to represent a value that may or may not be present.

That it's not optional in the resume payload is represented in that payload type, where we can't pass an Optional field in as the values.

CoreyShupe: Gotcha, that makes sense. I'll change it back to Optional

ianagbip1oti (approved):

